PR #2, opened 2019-07-16
CMakeLists file updated

SatjaSivcev:
- Version requirement removed.
- Copying dependencies made universal.
- This file can now be used with Visual Studio and QT creator.

### DIFF
--- a/Zivid/CMakeLists.txt
+++ b/Zivid/CMakeLists.txt
@@ -4,8 +4,18 @@ cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
 
 set(CMAKE_CXX_STANDARD 11)
 
-find_package(Zivid COMPONENTS Core 1.2.1 REQUIRED)
+find_package(Zivid COMPONENTS Core REQUIRED)
     
+add_custom_target(
+    SampleDependencies
+    COMMAND ${CMAKE_COMMAND} -E copy_directory
+            $ENV{ZIVID_INSTALL_FOLDER}/$<$<CONFIG:Debug>:bin_debug>$<$<CONFIG:Release>:bin>/plugins/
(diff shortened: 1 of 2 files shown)


nedrebo:
You should add relevant reviewers to your PRs to draw attention. Then it will show up in the reviewers dashboards.

I would have added me, but especially @apartridge since he wrote the initial code and he did not participate in writing this patch. Me reviewing this would be less valuable since I wrote most of this patch.

nedrebo:
The other top level CMakeLists.txt also needs updating.

As discussed before, I think you should consider having only one top level CMakeLists.txt. That would reduce code duplication.

ZividPCL/CMakeLists.txt
```diff
@@ -22,14 +32,7 @@ foreach(SAMPLE_NAME ${SAMPLE_VIS_NAMES})
 		Zivid::Vis3D
 		${PCL_LIBRARIES}
 	)
+	
+	add_dependencies(${SAMPLE_NAMES} SampleDependencies)
```
knatten: This should say `SAMPLE_NAME` instead of `SAMPLE_NAMES`.

Zivid/CMakeLists.txt
```diff
@@ -17,13 +27,7 @@ foreach(SAMPLE_NAME ${SAMPLE_NAMES})
 		Zivid::Vis3D
 	)
 
+    add_dependencies(${SAMPLE_NAMES} SampleDependencies)
```
knatten: This should say `SAMPLE_NAME` instead of `SAMPLE_NAMES`.

knatten:
Talked to @nedrebo now, let's merge this as is, and then solve the thing with a common CMakeLists as a separate issue. Created #7 for that.

Zivid/CMakeLists.txt
knatten: Remember to do this on the downsample branch as well btw.

Issue #3, opened 2019-07-16
Set up CI for Linux and Windows


nedrebo:
Set up a CI for building on the supported platforms (W10, U16, U18).

We use Azure Pipelines for CI.

Refer to https://github.com/zivid/zivid-python for how to do it, especially look into https://github.com/zivid/zivid-python/blob/master/azure-pipelines.yml. Building this repository should be very similar, just simpler since all the Python things can be skipped.

Ask @nedrebo for help enabling this in Azure and @apartridge or @trym-b for code changes.

nedrebo:
This is only partially fixed by #13. Still missing building with warnings, clang-tidy and coverage on Windows and Ubuntu 16.04 as well.

nedrebo:
@knatten Please re-open, seems I cannot do it (https://stackoverflow.com/a/21333938).

nedrebo:
This issue seems to have been split into #15, #16 and #17. No need for re-opening.

knatten:
Yeah I split it into more focused issues. Should have made a note about it here.

PR #4, opened 2019-07-30
Downsample

SatjaSivcev:


SatjaSivcev:
@knatten Can you please review this code? Be strict, let's make it solid.
For the python-samples we have added some automatic checking on Travis (black, flake8, pylint). Should we add something here as well?

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	catch (const std::exception &e)
+	{
+		std::cerr << "Error: " << Zivid::toString(e) << std::endl;
+		return EXIT_FAILURE;
+	}
+}
+
+Eigen::MatrixXf sumline(Eigen::MatrixXf& matrix, int downSamplingFactor)
```
knatten: In general I prefer to have functions in "reading order". So it's good that you have `main` on top, as that's the entry point. Then I'd continue with `downsample`, as that's the one being called from `main`. Then I'd continue with the functions called by `downsample`, and then the ones called from those again etc.

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+	return reshapedMatrix.colwise().sum().reshaped((int)matrix.rows() / downSamplingFactor, (int)matrix.cols()).transpose();
+}
+
+Eigen::MatrixXf gridsum(Eigen::MatrixXf& r, int downSamplingFactor)
+{
+	return sumline(sumline(r, downSamplingFactor), downSamplingFactor);
+}
+
+float myround(float x)
```
knatten: What's this function for? Would it make the code simpler if you remove it and use `std::round` directly instead?

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+	return std::round(x);
+}
+
+float isnanmask(float x)
```
knatten: What's this function for? Would it make the code simpler if you remove it and replace all usages with `!std::isnan` directly instead?

If you do keep this function, the function body can be simplified to `return !std::isnan(x)`.

I also find it surpising that a `is...` function returns `float`, as `is...` sounds like predicate that should return `bool`.

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+Zivid::PointCloud downsample(Zivid::PointCloud& pointCloud, int downSamplingFactor)
+{
+	/*
+	Function for downsampling a Zivid point cloud. The downsampling factor represents the denominator of a fraction that represents the size of the downsampled point cloud relative to the original point cloud, e.g. 2 - one-half,  3 - one-third, 4 one-quarter, etc.
+	*/
+
+	if ((downSamplingFactor % 2 != 0) || (pointCloud.height() % downSamplingFactor) || (pointCloud.width() % downSamplingFactor))
+	{
+		throw std::invalid_argument("Downsampling factor has to have one of the following values: 2, 3, 4, 5, 6.");
```
knatten: This does not seem to be what the `if` checks for? It looks like it checks that the factor is a multiple of two, and that the point cloud height and width are multiples of the factor.

The exception message has to reflect the actual requirement.

Does it really have to be a multiple of two? If so, I suggest splitting the `if` in two checks. One for checking whether it's a multiple of two, with an exception message saying just that, and one for checking that the dimensions are multiples of the factor, with an exception message which includes the dimensions and the factor.

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+			b(i, j) = pointCloud(i, j).blue();
+			contrast(i, j) = pointCloud(i, j).contrast;
+		}
+	}
+
+	Eigen::MatrixXf rf = r.template cast<float>();
+	Eigen::MatrixXf gf = g.template cast<float>();
+	Eigen::MatrixXf bf = b.template cast<float>();
+
+	std::function<float(float)> round_wrap = myround;
```
knatten: Do you need this wrapper?

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+	Eigen::MatrixXi redDownsampled = ((gridsum(rf, downSamplingFactor) / (downSamplingFactor*downSamplingFactor)).unaryExpr(round_wrap)).template cast<int>();
+	Eigen::MatrixXi greenDownsampled = ((gridsum(gf, downSamplingFactor) / (downSamplingFactor*downSamplingFactor)).unaryExpr(round_wrap)).template cast<int>();
+	Eigen::MatrixXi blueDownsampled = ((gridsum(bf, downSamplingFactor) / (downSamplingFactor*downSamplingFactor)).unaryExpr(round_wrap)).template cast<int>();
+
+	std::function<float(float)> is_nan_mask_wrap = isnanmask;
```
knatten: Do you need these wrappers?

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+
+	Eigen::MatrixXf rf = r.template cast<float>();
+	Eigen::MatrixXf gf = g.template cast<float>();
+	Eigen::MatrixXf bf = b.template cast<float>();
+
+	std::function<float(float)> round_wrap = myround;
+
+	Eigen::MatrixXi redDownsampled = ((gridsum(rf, downSamplingFactor) / (downSamplingFactor*downSamplingFactor)).unaryExpr(round_wrap)).template cast<int>();
```
knatten: Maybe you could extract this line and the `Eigen::MatrixXf rf = r.template cast<float>();` line into a function which takes takes in the `MatrixXi` and the downsampling factor and returns the new `MatrixXi`?

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+	Eigen::MatrixXi greenDownsampled = ((gridsum(gf, downSamplingFactor) / (downSamplingFactor*downSamplingFactor)).unaryExpr(round_wrap)).template cast<int>();
+	Eigen::MatrixXi blueDownsampled = ((gridsum(bf, downSamplingFactor) / (downSamplingFactor*downSamplingFactor)).unaryExpr(round_wrap)).template cast<int>();
+
+	std::function<float(float)> is_nan_mask_wrap = isnanmask;
+	std::function<float(float)> nan_to_zero_wrap = nantozero;
+
+	Eigen::MatrixXf contrastNulled = (z.unaryExpr(is_nan_mask_wrap)).cwiseProduct(contrast.unaryExpr(nan_to_zero_wrap));
```
knatten: It would probably be simpler to delete `is_nan_mask_wrap` and `isnanmask`, and use a lambda here instead. Something like:

```c++
z.unaryExpr([](float f){ return !std::isnan(f); })
```

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+	Eigen::MatrixXf zDownsampled = gridsum(zContrasted, downSamplingFactor).cwiseQuotient(gridsum(contrastNulled, downSamplingFactor));
+
+	Zivid::PointCloud pointCloudDownsampled(redDownsampled.rows(), redDownsampled.cols());
+
+	for (int i = 0; i < redDownsampled.rows(); i++)
+	{
+		for (int j = 0; j < redDownsampled.cols(); j++)
+		{
+			pointCloudDownsampled(i, j).setRgb(redDownsampled(i, j), greenDownsampled(i, j), blueDownsampled(i, j));
+			//pointCloudDownsampled(i, j).setContrast(contrastDownsampled(i,j));
```
knatten: Commented out code, can it be deleted?

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+	{
+		return 0;
+	}
+	else
+	{
+		return x;
+	}
+}
+
+Zivid::PointCloud downsample(Zivid::PointCloud& pointCloud, int downSamplingFactor)
```
knatten: Can `pointCloud` be `const`?

knatten:
By the way, if you want to we can sit together and make the requested changes.

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+		auto pointCloud = frame.getPointCloud();
+		auto downSamplingFactor = 4;
+
+		Zivid::PointCloud pointCloudDownsampled(pointCloud.height()/downSamplingFactor, pointCloud.width()/downSamplingFactor);
+
+		pointCloudDownsampled = downsample(pointCloud, downSamplingFactor);
+
+		std::cout << "Setting up visualization" << std::endl;
+		{
```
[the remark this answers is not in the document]
SatjaSivcev: Then I need to use Zivid::Application zivid; and Zivid::CloudVisualizer vis; in the function?

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+int main()
+{
+	try
+	{
+		Zivid::Application zivid;
+
+		std::string Filename = "Zivid3D.zdf";
+		std::cout << "Reading " << Filename << " point cloud" << std::endl;
+		Zivid::Frame frame = Zivid::Frame(Filename);
```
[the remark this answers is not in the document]
SatjaSivcev: Changed to Zivid::Frame frame(Filename);

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+float nantozero(float);
+Zivid::PointCloud downsample(Zivid::PointCloud&, int);
+
+int main()
+{
+	try
+	{
+		Zivid::Application zivid;
+
+		std::string Filename = "Zivid3D.zdf";
```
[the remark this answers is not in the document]
SatjaSivcev: done

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,196 @@
...
+		std::cout << "Reading " << Filename << " point cloud" << std::endl;
+		Zivid::Frame frame = Zivid::Frame(Filename);
+
+		auto pointCloud = frame.getPointCloud();
+		auto downSamplingFactor = 4;
+
+		Zivid::PointCloud pointCloudDownsampled(pointCloud.height()/downSamplingFactor, pointCloud.width()/downSamplingFactor);
```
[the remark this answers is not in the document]
SatjaSivcev: done

Downsample/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+            ${CMAKE_CURRENT_SOURCE_DIR}/../Zivid3D.zdf
+            ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}
+)
+
+#find_package(Eigen3 REQUIRED)
+SET( EIGEN3_INCLUDE_DIR "$ENV{EIGEN3_INCLUDE_DIR}" )
+IF( NOT EIGEN3_INCLUDE_DIR )
+    MESSAGE( FATAL_ERROR "Please point the environment variable EIGEN3_INCLUDE_DIR to the include directory of your Eigen3 installation.")
+ENDIF()
+INCLUDE_DIRECTORIES ( "${EIGEN3_INCLUDE_DIR}" )
```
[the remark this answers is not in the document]
SatjaSivcev: I couldn't make this work so I had to manually import Eigen headers path. I think my cmake has the worng path and I don't know how to change it.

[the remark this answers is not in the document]
SatjaSivcev: I tried to build one version and failed. Then I just downloaded another version which I am using now.

Downsample/Downsample.cpp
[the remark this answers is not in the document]
knatten: I think you need to create the `CloudVisualizer` inside the function, just as you create it inside the scope now. At least if I'm correct that you're recreating it since you can't reuse it?

And then you would need to take `zivid` by reference as a parameter to the function.

Downsample/Downsample.cpp
SatjaSivcev: I don't know. But this doesn't work:
Eigen::MatrixXf contrastNulled = (z.unaryExpr(isnanmask)).cwiseProduct(contrast.unaryExpr(nantozero));

Maybe coding is wrong?

Downsample/Downsample.cpp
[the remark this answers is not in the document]
SatjaSivcev: done

Downsample/CMakeLists.txt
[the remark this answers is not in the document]
SatjaSivcev: No need to find the package because the tutorial will have:
cmake DEIGEN3_INCLUDE_DIR="path_to_eigen" ..
It will also tell exactly which version is needed and where to clone it.

Downsample/Downsample.cpp
SatjaSivcev: Some we left some we don't need. Functor is needed however.

Downsample/Downsample.cpp
SatjaSivcev: done

Downsample/Downsample.cpp
SatjaSivcev: done

Downsample/Downsample.cpp
SatjaSivcev: done

Downsample/Downsample.cpp
SatjaSivcev: done

nedrebo:
> @knatten Can you please review this code? Be strict, let's make it solid.
> For the python-samples we have added some automatic checking on Travis (black, flake8, pylint). Should we add something here as well?

This is covered by #3, yes we should do that.

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	return pointCloudDownsampled;
+}
+
+Eigen::MatrixXi downsampleandround(const Eigen::MatrixXi& matrixi, int downsamplingFactor)
+{
+	Eigen::MatrixXf matrixf = matrixi.template cast<float>();
+
+	std::function<float(float)> round_wrap = [](float f) {return std::round(f); };
```
knatten: Nitpick, call this `functor` instead of `wrap`, since we did that for the other ones.

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,167 @@
...
+*/
+
+#include <Zivid/CloudVisualizer.h>
+#include <Zivid/Zivid.h>
+#include <Eigen/Dense>
+#include <Eigen/Core>
+
+#include <iostream>
+Zivid::PointCloud downsample(const Zivid::PointCloud&, int);
+Eigen::MatrixXi downsampleandround(const Eigen::MatrixXi&, int);
```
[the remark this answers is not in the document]
SatjaSivcev: done

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,167 @@
...
+Eigen::MatrixXf gridsum(const Eigen::MatrixXf&, int);
+Eigen::MatrixXf sumline(const Eigen::MatrixXf&, int);
+float nantozero(float);
+void visualizepointcloud(const Zivid::PointCloud&, Zivid::Application&);
```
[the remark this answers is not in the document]
SatjaSivcev: Done. I camel cased all functions.

Downsample/Downsample.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+Zivid::PointCloud downsample(const Zivid::PointCloud& pointCloud, int downsamplingFactor)
+{
+	/*
+	Function for downsampling a Zivid point cloud. The downsampling factor represents the denominator of a fraction that represents the size of the downsampled point cloud relative to the original point cloud, e.g. 2 - one-half,  3 - one-third, 4 one-quarter, etc.
+	*/
+
+	if ((pointCloud.height() % downsamplingFactor) || (pointCloud.width() % downsamplingFactor))
+	{
+		throw std::invalid_argument("Downsampling factor (" + std::to_string(downsamplingFactor) + ") has to a factor of the width (" + std::to_string(pointCloud.width()) + ") and height (" + std::to_string(pointCloud.height()) + ") of the input point cloud.");
```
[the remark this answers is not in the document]
SatjaSivcev: What do you mean by splitting it?

Downsample/Downsample.cpp
SatjaSivcev: done.

Downsample/Downsample.cpp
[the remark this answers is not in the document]
knatten: Just split it into several lines. I ran it through the default `clang-format` which suggests this:
```c++
    throw std::invalid_argument(
        "Downsampling factor (" + std::to_string(downsamplingFactor) +
        ") has to a factor of the width (" +
        std::to_string(pointCloud.width()) + ") and height (" +
        std::to_string(pointCloud.height()) + ") of the input point cloud.");
```
But this would also be ok given the current overall line length in that file:
```c++
    throw std::invalid_argument(
        "Downsampling factor (" + std::to_string(downsamplingFactor) + ") has to a factor of the width (" +
        std::to_string(pointCloud.width()) + ") and height (" + std::to_string(pointCloud.height()) + ") of the input point cloud."
        );
```
This will sort itself out as soon as we add automatic formatting. I'll lobby for doing that sooner rather than later in our upcoming sprint planning.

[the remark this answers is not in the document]
SatjaSivcev: Done.

SatjaSivcev:
Ready for the next round.

knatten:
Since we usually use `camelCase` everywhere else, I think we should use `camelCase` instead of `snake_case` here too. So `downsampleAndRound` instead of `downsample_and_round` etc.

SatjaSivcev:
Ready for the next round

knatten:
Nicely done! Approved. Want help cleaning up the git history before we merge?

knatten:
Reverted this merge now, as agreed with @SatjaSivcev and @nedrebo. Have a few things we need to fix, then we'll make a new PR.

PR #5, opened 2019-07-31
Add link to readme

SatjaSivcev:
README file links updated and fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,29 +3,31 @@
 
 This repository contains additional **C++** code samples for **Zivid**.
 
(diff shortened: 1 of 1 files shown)


SatjaSivcev:
Relative links don't work from Confluence page so we need to keep them full pointing to the master.

knatten:
Right, so the README is imported into Confluence in some way? Can you show me how that works, maybe we could work around that problem?

knatten:
Haven't figured out how to use relative links yet, but will merge it for now. Added #8 to switch to relative links.

PR #6, opened 2019-08-02
Set up CI with Azure Pipelines

zividbuilder:


knatten:
This PR was apparently made automatically when Azure pipelines were set up for the repo (see #3). We should remove the dummy steps and add some actual CI in here before merging (again see #3).

knatten:
Replaced by #13

PR #12, opened 2019-08-16
Use clang format

knatten:
This is the first step for enabling CI on this repository (See #3).

I think it makes sense to merge this PR separately even if the full CI setup is not complete, to make sure we use proper formatting sooner rather than later. This will avoid adding even more code with inconsistent formatting until the CI setup is complete.


SatjaSivcev:
Ready for merge.

PR #21, opened 2019-08-29
Add OpenCV CaptureUndistortRGB sample

SatjaSivcev:
The sample undistorts an RGB image from a ZDF point cloud using Zivid camera intrinsics.

### DIFF
--- a/ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
+++ b/ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
@@ -1,0 +1,139 @@
+/*
+Undistort an RGB image from a ZDF point cloud using Zivid camera intrinsics.
+*/
+
+#include <Zivid/CloudVisualizer.h>
+#include <Zivid/Zivid.h>
+
+#include <opencv2/calib3d/calib3d.hpp>
+#include <opencv2/core/core.hpp>
+#include <opencv2/highgui/highgui.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+
+#include <iostream>
+
+cv::Mat pointCloudToRGB(const Zivid::PointCloud &);
+std::tuple<cv::Mat, cv::Mat> reformatCameraIntrinsics(const Zivid::CameraIntrinsics &);
+void displayRGB(const cv::Mat &, const std::string &);
+
+int main()
+{
+    try
+    {
+        Zivid::Application zivid;
+
+        std::cout << "Connecting to the camera" << std::endl;
+        auto camera = zivid.connectCamera();
+
+        std::cout << "Configuring the camera settings" << std::endl;
+        camera << Zivid::Settings::Iris{ 21 } << Zivid::Settings::ExposureTime{ std::chrono::microseconds{ 20000 } }
+               << Zivid::Settings::Gain{ 1 } << Zivid::Settings::Brightness{ 1.0 };
+
+        std::cout << "Capturing a frame" << std::endl;
+        const auto frame = camera.capture();
+
+        std::cout << "Setting up visualization" << std::endl;
+        Zivid::CloudVisualizer vis;
+        zivid.setDefaultComputeDevice(vis.computeDevice());
+
+        std::cout << "Displaying the point cloud" << std::endl;
+        vis.showMaximized();
+        vis.show(frame);
+        vis.resetToFit();
+
+        std::cout << "Running the visualizer. Blocking until the window closes" << std::endl;
+        vis.run();
+
+        std::cout << "Converting ZDF point cloud to OpenCV format" << std::endl;
+
+        const auto pointCloud = frame.getPointCloud();
+        const auto rgb = pointCloudToRGB(pointCloud);
+
+        std::cout << "Undistorting the RGB image" << std::endl;
+
+        const auto cameraIntrinsticsCV = reformatCameraIntrinsics(camera.intrinsics());
+        const auto distortionCoefficients = std::get<0>(cameraIntrinsticsCV);
+        const auto cameraMatrix = std::get<1>(cameraIntrinsticsCV);
+
+        const auto size = rgb.size();
+        const auto optimalCameraMatrix =
+            cv::getOptimalNewCameraMatrix(cameraMatrix, distortionCoefficients, size, 1, size);
+
+        cv::Mat rgbUndistorted;
+        cv::Mat rgbUndistortedFull;
+
+        cv::undistort(rgb, rgbUndistorted, cameraMatrix, distortionCoefficients);
+        cv::undistort(rgb, rgbUndistortedFull, cameraMatrix, distortionCoefficients, optimalCameraMatrix);
+
+        std::cout << "Displaying and saving the RGB image" << std::endl;
+
+        displayRGB(rgb, "Distorted RGB image");
+        cv::imwrite("Distorted RGB image.jpg", rgb);
+
+        std::cout << "Displaying and saving the undistorted RGB image" << std::endl;
+
+        displayRGB(rgbUndistorted, "Undistorted RGB image");
+        cv::imwrite("Undistorted RGB image.jpg", rgbUndistorted);
+
+        std::cout << "Displaying and saving the Undistorted RGB image - full" << std::endl;
+
+        displayRGB(rgbUndistortedFull, "Undistorted RGB image - full");
+        cv::imwrite("Undistorted RGB image - full.jpg", rgbUndistorted);
+    }
+    catch(const std::exception &e)
+    {
+        std::cerr << "Error: " << Zivid::toString(e) << std::endl;
+        return EXIT_FAILURE;
+    }
+}
+
+cv::Mat pointCloudToRGB(const Zivid::PointCloud &pointCloud)
+{
+    cv::Mat rgb(static_cast<int>(pointCloud.height()),
+                static_cast<int>(pointCloud.width()),
+                CV_8UC3,
+                cv::Scalar(0, 0, 0));
+
+    const auto height = pointCloud.height();
+    const auto width = pointCloud.width();
+
+    for(int i = 0; i < height; i++)
+    {
+        for(int j = 0; j < width; j++)
+        {
+            cv::Vec3b &color = rgb.at<cv::Vec3b>(i, j);
+            color[0] = pointCloud(i, j).blue();
+            color[1] = pointCloud(i, j).green();
+            color[2] = pointCloud(i, j).red();
+        }
+    }
+
+    return rgb;
+}
+
+std::tuple<cv::Mat, cv::Mat> reformatCameraIntrinsics(const Zivid::CameraIntrinsics &cameraIntrinsics)
+{
+    cv::Mat distortionCoefficients(cv::Size(1, 5), CV_64FC1, cv::Scalar(0));
+    cv::Mat cameraMatrix(cv::Size(3, 3), CV_64FC1, cv::Scalar(0));
+
+    distortionCoefficients.at<double>(0, 0) = cameraIntrinsics.distortion().k1().value();
+    distortionCoefficients.at<double>(0, 1) = cameraIntrinsics.distortion().k2().value();
+    distortionCoefficients.at<double>(0, 2) = cameraIntrinsics.distortion().p1().value();
+    distortionCoefficients.at<double>(0, 3) = cameraIntrinsics.distortion().p2().value();
+    distortionCoefficients.at<double>(0, 4) = cameraIntrinsics.distortion().k3().value();
+
+    cameraMatrix.at<double>(0, 0) = cameraIntrinsics.cameraMatrix().fx().value();
+    cameraMatrix.at<double>(0, 2) = cameraIntrinsics.cameraMatrix().cx().value();
+    cameraMatrix.at<double>(1, 1) = cameraIntrinsics.cameraMatrix().fy().value();
+    cameraMatrix.at<double>(1, 2) = cameraIntrinsics.cameraMatrix().cy().value();
+    cameraMatrix.at<double>(2, 2) = 1;
+
+    return std::make_tuple(distortionCoefficients, cameraMatrix);
+}
+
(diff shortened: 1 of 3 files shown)


knatten:
I think this looks very good in general! The sample is a bit long though, what do you think about extracting some functions? You could for instance extract functions for creating the `rgb` matrix, `distortionCoefficients` and `cameraMatrix`. That would make `main` easier to read imo, and it would make it more obvious what those blocks of code do.

You could maybe also extract a function for displaying rgb images, since that code is duplicated three times.

On the other hand, there is a certain value in having a sample just be one block of code which can be read from top to bottom.

I'm happy to approve it either way. What do you think?

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,129 @@
...
+    distortionCoefficients.at<double>(0, 4) = camera.intrinsics().distortion().k3().value();
+
+    cameraMatrix.at<double>(0, 0) = camera.intrinsics().cameraMatrix().fx().value();
+    cameraMatrix.at<double>(0, 2) = camera.intrinsics().cameraMatrix().cx().value();
+    cameraMatrix.at<double>(1, 1) = camera.intrinsics().cameraMatrix().fy().value();
+    cameraMatrix.at<double>(1, 2) = camera.intrinsics().cameraMatrix().cy().value();
+    cameraMatrix.at<double>(2, 2) = 1;
+}
+
+void displayRGB(const cv::Mat &rgb, const std::string rgbName)
```
knatten: `rgbName` can be a reference. It won't matter performancewise, but it's just less surprising to read.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,129 @@
...
+        cv::imwrite("Undistorted RGB image - full.jpg", rgbUndistorted);
+    }
+    catch(const std::exception &e)
+    {
+        std::cerr << "Error: " << Zivid::toString(e) << std::endl;
+        return EXIT_FAILURE;
+    }
+}
+
+void pointCloudToRGB(const Zivid::PointCloud &pointCloud, cv::Mat &rgb)
```
knatten: We should avoid out parameters like `cv::Mat &rgb`, and instead use return values. Move the initialization of `rgb` inside the `pointCloudToRGB` function, and then return it by value. (The resulting `rgb` variable in `main()` should then also be made `const`)
The C++ Core Guidelines have a [section about this](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f20-for-out-output-values-prefer-return-values-to-output-parameters), although only the first line is really relevant in this case.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,129 @@
...
+        {
+            cv::Vec3b &color = rgb.at<cv::Vec3b>(i, j);
+            color[0] = pointCloud(i, j).blue();
+            color[1] = pointCloud(i, j).green();
+            color[2] = pointCloud(i, j).red();
+        }
+    }
+}
+
+void getCameraIntrinsics(const Zivid::Camera &camera, cv::Mat &distortionCoefficients, cv::Mat &cameraMatrix)
```
knatten: We should avoid out parameters here too, and initialize the variables inside the function and then return them.

One alternative is to return both values using a struct or tuple, like the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f21-to-return-multiple-out-values-prefer-returning-a-struct-or-tuple) suggest when having to  return several values.

An alternative would be to split this in two functions, one for creating `distortionCoefficients` and one for creating `cameraMatrix`, then you only need one return value for each function.

In either case, the resulting variables in `main()` should be made `const`.

You can also consider passing `camera.intrinsics()` or `camera.intrinsics().distortion()` into the function(s) instead of passing `camera`. I think this will make the code a bit cleaner since you don't have to keep calling all those functions.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,129 @@
...
+        std::cout << "Undistorting the RGB image" << std::endl;
+
+        cv::Size size = rgb.size();
+
+        cv::Mat rgbUndistorted;
+        cv::Mat rgbUndistortedFull;
+
+        cv::Mat distortionCoefficients(cv::Size(1, 5), CV_64FC1, cv::Scalar(0));
+        cv::Mat cameraMatrix(cv::Size(3, 3), CV_64FC1, cv::Scalar(0));
+        cv::Mat optimalCameraMatrix;
```
knatten: This is now first default initialized on this line, and then re-initialized a few lines below. You can do this on one single line instead, and avoiding un-used temporary objects. That will make the code clearer, and reduce the risk of someone using an improperly initialized variable. It also allows you to make it `const`.

C++ Core Guidelines: [ES.22: Don't declare a variable until you have a value to initialize it with](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es22-dont-declare-a-variable-until-you-have-a-value-to-initialize-it-with)

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+        std::cout << "Converting ZDF point cloud to OpenCV format" << std::endl;
+
+        const auto pointCloud = frame.getPointCloud();
+        cv::Mat rgb((int)pointCloud.height(), (int)pointCloud.width(), CV_8UC3, cv::Scalar(0, 0, 0));
+        pointCloudToRGB(pointCloud, rgb);
+
+        std::cout << "Undistorting the RGB image" << std::endl;
+
+        cv::Size size = rgb.size();
```
knatten: I suggest moving this variable closer to where it's used. Then the user doesn't have to keep it in mind for longer than necessary when reading the code, or wonder what it's going to be used for.

C++ Core Guidelines: [ES.21: Don't introduce a variable (or constant) before you need to use it](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es21-dont-introduce-a-variable-or-constant-before-you-need-to-use-it).

The same goes for `rgbUndistorted` and `rgbUndistortedFull` below.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,129 @@
...
+        vis.show(frame);
+        vis.resetToFit();
+
+        std::cout << "Running the visualizer. Blocking until the window closes" << std::endl;
+        vis.run();
+
+        std::cout << "Converting ZDF point cloud to OpenCV format" << std::endl;
+
+        const auto pointCloud = frame.getPointCloud();
+        cv::Mat rgb((int)pointCloud.height(), (int)pointCloud.width(), CV_8UC3, cv::Scalar(0, 0, 0));
```
knatten: Avoid C-style casts, and use a `static_cast` instead.

C++ Core Guidelines: [ES.49: If you must use a cast, use a named cast](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es49-if-you-must-use-a-cast-use-a-named-cast)

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,129 @@
...
+    catch(const std::exception &e)
+    {
+        std::cerr << "Error: " << Zivid::toString(e) << std::endl;
+        return EXIT_FAILURE;
+    }
+}
+
+void pointCloudToRGB(const Zivid::PointCloud &pointCloud, cv::Mat &rgb)
+{
+    for(int i = 0; i < pointCloud.height(); i++)
```
knatten: Don't call `height()` and `width()` for each iteration of the for loop. The compiler will probably not be able to optimize it by moving it outside the for loop, so this might result in a million extra function calls.

I did an experiment here https://godbolt.org/z/GcurlR and even with -O2 it calls the `width()` function on every iteration.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done. I think.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done with a tuple and passing camera.intrinsics(). This is fun.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: I think this is done if I understand correctly what the problem is.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done

SatjaSivcev:
Ready for the next round. Heads up: I may have used const too much.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
knatten: Nice!

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,139 @@
...
+            color[0] = pointCloud(i, j).blue();
+            color[1] = pointCloud(i, j).green();
+            color[2] = pointCloud(i, j).red();
+        }
+    }
+
+    return rgb;
+}
+
+std::tuple<cv::Mat, cv::Mat> getCameraIntrinsics(const Zivid::CameraIntrinsics &cameraIntrinsics)
```
knatten: Can we think of a better name? We've kind of already gotten the `CameraIntrinsic`s, and are passing them to the function. So this function is more about converting it to to OpenCV formats.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+cv::Mat pointCloudToRGB(const Zivid::PointCloud &pointCloud)
+{
+    cv::Mat rgb(static_cast<int>(pointCloud.height()),
+                static_cast<int>(pointCloud.width()),
+                CV_8UC3,
+                cv::Scalar(0, 0, 0));
+
+    auto height = pointCloud.height();
```
knatten: Nitpick: These can also be const

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+        std::cout << "Converting ZDF point cloud to OpenCV format" << std::endl;
+
+        const auto pointCloud = frame.getPointCloud();
+        const auto rgb = pointCloudToRGB(pointCloud);
+
+        std::cout << "Undistorting the RGB image" << std::endl;
+
+        const auto cameraIntrinsticsCV = getCameraIntrinsics(camera.intrinsics());
+        const cv::Mat distortionCoefficients = std::get<0>(cameraIntrinsticsCV);
```
knatten: For consistency, maybe use `auto` for these four variables too?

knatten:
> Heads up: I may have used const too much.

One can never use `const` too much! <3

knatten:
Looks very good now! Just have some small nitpicks and this will be great.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: reformatCameraIntrinsics? I couldn't think of anything better. cameraIntrinsicsToOpenCV didn't look nice.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
SatjaSivcev: Done, if we assume the same four.

SatjaSivcev:
Ready for the next round.

knatten:
Looking great! We should merge #23 first, and then rebase this PR on master. That will require updating the new root `CMakeLists.txt`. I can help with both of those if you want.

Issue #15, opened 2019-08-16
Set up CI that builds all examples on Linux


knatten:
Refer to https://github.com/zivid/zivid-python and do it the same way

nedrebo:
Zivid officially support Ubuntu 16.04 and 18.04, both those should be covered.

However, the additional distributions tested against in zivid-python is probably overkill and for this and will only result in maintenance overhead.

knatten:
It's probably a good idea to also check that samples with missing dependencies build when the dependency is turned off. For instance setting `OpenCV_DIR` to a non existing directory and `USE_OPENCV=OFF` should work.
This will reveal if we add a sample to `CMakeLists.txt` without adding it to the proper dependency lists.

(Idea came from https://github.com/zivid/cpp-extra-samples/pull/23#discussion_r325063844)

knatten:
Should build on both Ubuntu 18.04 and 16.04.

knatten:
We're building on Linux in CI. I've added #42 and #43 to improve further.

PR #23, opened 2019-09-17
One CMakeLists.txt for all samples

knatten:
This makes it easier for users, they now only have to configure one
CMake project to build all samples. They can also open just one
directory to open all samples in their IDE.

Using one CMakeLists.txt also gets rid of a lot of code duplication.

Many of the samples depend on third party libraries, in particular Eigen
3, OpenCV or PCL. The user might not be interested in installing all of
these dependencies to get some samples to work, so we included options
USE_EIGEN3 etc. to allow turning samples with certain dependencies on or
off. The user can then choose between providing a third party library,
or turning off samples depending on that library.

An alternative approach was discussed, where all samples with third
party dependencies were turned off by default, and the user would have
to explicitly turn them on. We decided against this, since we are afraid
they'll then simply never notice these samples at all.

This approach with one CMakeLists.txt also solves the potential future
problem of how to handle samples which have more than one dependency. In
this case the previous approach with one directory per dependency would
fall apart.

Finally, we also introduce an option to turn off all samples which
depend on Zivid::Vis3D, as this is currently hard to get to build on
Linux. We'll hopefully solve that problem soon, but in the mean time
it's convenient to be able to skip those samples.


CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+        set(DEPENDENCY_LIST "${DEPENDENCY_NAME}_DEPENDING")
+        if(${SAMPLE_NAME} IN_LIST ${DEPENDENCY_LIST})
+            message("  - ${SAMPLE}")
+            list(REMOVE_ITEM SAMPLES ${SAMPLE})
+        endif()
+    endforeach()
+endmacro()
+
+if(USE_EIGEN3)
+    if(NOT DEFINED EIGEN3_INCLUDE_DIR)
```
nedrebo: Why don't you wait with find_package(.. REQUIRED ..) until here? Then you would get the official error message to the consumer and you dont have to invent your own, that needs to be kept in sync with Eigen's config/find script.

CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+                ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/plugins
+    )
+endif()
+
+message(STATUS "All samples: ${SAMPLES}")
+
+foreach(SAMPLE ${SAMPLES})
+    get_filename_component(SAMPLE_NAME ${SAMPLE} NAME)
+
+	add_executable(${SAMPLE_NAME} ${SAMPLE}/${SAMPLE_NAME}.cpp)
```
nedrebo: Indentation error

CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+    Zivid/ReadZDF
+    Zivid/ZDF2PLY
+    ZividPCL/ReadPCLVis3D
+    ZividPCL/CaptureWritePCLVis3D
+    ZividPCL/CaptureFromFileWritePCLVis3D
+    ZividPCL/ZDF2PCD
+    ZividOpenCV/ZDF2OpenCV
+    Downsample)
+
+set(Eigen3_DEPENDING Downsample)
```
nedrebo: There is no (guaranteed) correlation between this lists and the directory convention used above? The list below is more flexible than what can be expressed in the folder structure. Where are new samples supposed to be placed? For example one that depends on OpenCV and PCL.

CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+    Zivid/CaptureSavePLY
+    Zivid/ConnectToSerialNumberCamera
+    Zivid/GetCameraIntrinsics
+    Zivid/ReadZDF
+    Zivid/ZDF2PLY
+    ZividPCL/ReadPCLVis3D
+    ZividPCL/CaptureWritePCLVis3D
+    ZividPCL/CaptureFromFileWritePCLVis3D
+    ZividPCL/ZDF2PCD
+    ZividOpenCV/ZDF2OpenCV
```
nedrebo: If you add a Sample here that depends on PCL and you forget to add it to the list below then the CI would still pass I guess.

I think the CI should have a list of deps and first build without deps, then rerun cmake and add one and one of the options and build during each iteration. This should add almost nothing to total build time as it will be incremental builds.

nedrebo:
Oh no, seems like I published my review after this was merged. I would still appreciate if the comments was answered or addressed. Sorry for the inconvenience @SatjaSivcev and @knatten.

CMakeLists.txt
knatten: I discussed that with @SatjaSivcev a few days ago and agreed it's outside the scope of this PR. With this PR in place, they are now free to change the folder structure as they like.

CMakeLists.txt
knatten: Eigen is not using `find_package`,  I assume you're referring to PCL and OpenCV? That's a good point.

CMakeLists.txt
knatten: Yeah that's a potential problem. I've made a note of it in #15 .

CMakeLists.txt
knatten: I agree that it's a good idea to move `find_package` inside `if(USE....)`. And I assume you're referring to `PCL` and `OpenCV` only, since `Eigen` is not found by a CMake script.

In that case: There's nothing in those error messages which needs to be kept in sync with config/find scripts. `OpenCV_DIR` and `PCL_DIR` are not inputs to the OpenCV and PCL config files, they are arguments to CMake to know where to *look for* those config files in the first place. So I don't think this is actually a problem.

There's also another very valuable thing in the custom error messages: Letting the user know they can turn the dependency off with `-DUSE_THETHING=OFF`. So I think we should keep those messages.

My suggestion then is to move `find_package` inside `if(USE_...)`, but don't add `REQUIRED`. You can then both see CMake's verbose message about not finding a config file, *and* our own message.

What do you think @nedrebo and @SatjaSivcev ?

CMakeLists.txt
knatten: Fixed in #26

CMakeLists.txt
knatten: Made a draft PR #26 so you can see what I'm proposing.

nedrebo: #26 looks good

knatten: Cool, I added a better description and un-drafted it now. Thanks for the feedback! :)

PR #25, opened 2019-09-17
Add building of samples in CI

knatten:
I'm currently only building a subset of the samples which was easy to
get to build. Still, this will ensure those samples build, that cmake
works etc. Then we can expand with the OpenCV and Eigen samples later.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,7 @@ jobs:
     matrix:
       Ubuntu1804:
         os: 'ubuntu:18.04'
-        command: './test.sh'
+        command: './lint.sh && ./build.sh'
   steps:
   - script: docker run
               --volume $PWD:/host
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,11 +6,11 @@ jobs:
     matrix:
       Ubuntu1804:
         os: 'ubuntu:18.04'
-        command: './lint.sh'
+        command: './test.sh'
   steps:
   - script: docker run
               --volume $PWD:/host
               --workdir /host/continuous-integration/linux
               $(os)
               bash -c "./setup.sh && $(command)"
-    displayName: 'Lint'
+    displayName: 'Test'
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -6,11 +6,11 @@ jobs:
     matrix:
       Ubuntu1804:
         os: 'ubuntu:18.04'
-        command: './lint.sh'
+        command: './test.sh'
```
nedrebo: Did you intend to remove lint.sh

knatten: Yes, I'm calling it from `test.sh`

nedrebo: We usually don't do that. zivid-python, zivid-ros and gcps runs multiple entry scripts in the CI file. Not a big difference, but then you can run individual scripts on-demand.

continuous-integration/linux/test.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT_DIR=$(realpath "$SCRIPT_DIR/../..")
+
+"$SCRIPT_DIR/lint.sh"
```
nedrebo: Missing || exit, but I propose to keep this out of this file. My proposal is to have 4 top level scripts called
setup.py
build.py
test.sh
lint.sh

azure-pipelines.yml
nedrebo: The idea is that command can be `build.sh && test.sh` for most OSs and then it can be `lint.sh && build.sh && test.sh` for the most up to date OS.

knatten:
Thanks for the review! I've moved `lint.sh` back out now. I also renamed `test.sh` to `build.sh` to better reflect what it's doing and be more similar to the other repos.

Suggest not porting `build.sh` to Python until we add Windows CI as I have a lot of other things on my plate right now.

knatten:
Bah, rebasing on master dismissed your approval @nedrebo. Can you please re-approve?

knatten:
Bump @nedrebo

PR #30, opened 2019-10-29
Size() constructor inverted row column

torbsorb:
If you run the original code in debug mode you will get an assertion failure. 
`CV_DbgAssert((unsigned)(i1 * DataType<_Tp>::channels) < (unsigned)(size.p[1] * channels()));`
This is because as the Mat::Mat constructor documentation puts it:

> size – 2D array size: Size(cols, rows) . In the Size() constructor, the number of rows and the number of columns go in the reverse order.


knatten:
@SatjaSivcev you know this sample best, wanna review?

PR #31, opened 2019-10-29
Add 2D API to CaptureUndistortRGB example

torbsorb:
The last two commits represent two alternative approaches to a combined example.
fe39a35: Use #define to select whether to use 2D or 3D APIs
9698d44: Use std::getline() to prompt user to select 2D or 3D (actual implementation defaults to 3D after 5 seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,9 @@ Check out our [tutorial on configuring and building these samples on Ubuntu](htt
 
 - [**ZividOpenCV**](https://github.com/zivid/cpp-extra-samples/tree/master/ZividOpenCV)
 	- [**ZDF2OpenCV**](https://github.com/zivid/cpp-extra-samples/blob/master/ZividOpenCV/ZDF2OpenCV/ZDF2OpenCV.cpp) - Import a ZDF point cloud and convert it to OpenCV format.
-	- [**CaptureUndistortRGB**](https://github.com/zivid/cpp-extra-samples/blob/master/ZividOpenCV/ZDF2OpenCV/CaptureUndistortRGB.cpp) - Undistort an RGB image from a ZDF point cloud using Zivid camera intrinsics.
+	- [**CaptureUndistortRGB**](https://github.com/zivid/cpp-extra-samples/blob/master/ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp) - Use Zivid camera intrinsics to undistort an RGB image. This example will prompt the user for whether to capture a 2D or a 3D image. In both instances it will operate on a 2D image. However, in the 3D case it will extract 2D image from a ZDF point cloud. The 2D variant is faster.
 	- **Dependencies:**
-		- [OpenCV](https://opencv.org/) version 4.0.1-dev or newer
+		- [OpenCV](https://opencv.org/) version 4.0.1 or newer
 
 - [**HandEyeCalibration**](https://github.com/zivid/cpp-extra-samples/tree/master/HandEyeCalibration)
 	- [**UtilizeEyeInHandCalibration**](https://github.com/zivid/cpp-extra-samples/blob/master/HandEyeCalibration/UtilizeEyeInHandCalibration/UtilizeEyeInHandCalibration.cpp) - Transform a 3D point from camera frame to robot base frame using hand-eye calibration matrix.
--- a/ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
+++ b/ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
@@ -1,5 +1,5 @@
 /*
-Undistort an RGB image from a ZDF point cloud using Zivid camera intrinsics.
+Undistort a BGR image from a ZDF point cloud using Zivid camera intrinsics.
 */
 
 #include <Zivid/CloudVisualizer.h>
@@ -12,9 +12,13 @@ Undistort an RGB image from a ZDF point cloud using Zivid camera intrinsics.
 
 #include <iostream>
 
-cv::Mat pointCloudToRGB(const Zivid::PointCloud &);
+cv::Mat pointCloudToBGR(const Zivid::PointCloud &);
+cv::Mat imageToBGR(const Zivid::Image<Zivid::RGBA8> &);
 std::tuple<cv::Mat, cv::Mat> reformatCameraIntrinsics(const Zivid::CameraIntrinsics &);
-void displayRGB(const cv::Mat &, const std::string &);
+void displayBGR(const cv::Mat &, const std::string &);
+std::string getInput(void);
+cv::Mat getImage2D(Zivid::Camera &camera);
+cv::Mat getImage3D(Zivid::Camera &camera, Zivid::Application &zivid);
 
 int main()
 {
@@ -25,60 +29,46 @@ int main()
         std::cout << "Connecting to the camera" << std::endl;
         auto camera = zivid.connectCamera();
 
-        std::cout << "Configuring the camera settings" << std::endl;
-        camera << Zivid::Settings::Iris{ 21 } << Zivid::Settings::ExposureTime{ std::chrono::microseconds{ 20000 } }
-               << Zivid::Settings::Gain{ 1 } << Zivid::Settings::Brightness{ 1.0 };
+        std::cout << "Enter \"2d\" or \"3d\" to select mode, then press Enter/Return to confirm" << std::endl;
+        const auto command = getInput();
+        bool use2D = false;
+        if(command == "2d" || command == "2D")
+        {
+            use2D = true;
+        }
 
-        std::cout << "Capturing a frame" << std::endl;
-        const auto frame = camera.capture();
+        const auto bgr = (use2D ? getImage2D(camera) : getImage3D(camera, zivid));
 
-        std::cout << "Setting up visualization" << std::endl;
-        Zivid::CloudVisualizer vis;
-        zivid.setDefaultComputeDevice(vis.computeDevice());
-
-        std::cout << "Displaying the point cloud" << std::endl;
-        vis.showMaximized();
-        vis.show(frame);
-        vis.resetToFit();
-
-        std::cout << "Running the visualizer. Blocking until the window closes" << std::endl;
-        vis.run();
-
-        std::cout << "Converting ZDF point cloud to OpenCV format" << std::endl;
-
-        const auto pointCloud = frame.getPointCloud();
-        const auto rgb = pointCloudToRGB(pointCloud);
-
-        std::cout << "Undistorting the RGB image" << std::endl;
+        std::cout << "Undistorting the BGR image" << std::endl;
 
         const auto cameraIntrinsticsCV = reformatCameraIntrinsics(camera.intrinsics());
         const auto distortionCoefficients = std::get<0>(cameraIntrinsticsCV);
         const auto cameraMatrix = std::get<1>(cameraIntrinsticsCV);
 
-        const auto size = rgb.size();
+        const auto size = bgr.size();
         const auto optimalCameraMatrix =
             cv::getOptimalNewCameraMatrix(cameraMatrix, distortionCoefficients, size, 1, size);
 
-        cv::Mat rgbUndistorted;
-        cv::Mat rgbUndistortedFull;
+        cv::Mat bgrUndistorted;
+        cv::Mat bgrUndistortedFull;
 
-        cv::undistort(rgb, rgbUndistorted, cameraMatrix, distortionCoefficients);
-        cv::undistort(rgb, rgbUndistortedFull, cameraMatrix, distortionCoefficients, optimalCameraMatrix);
+        cv::undistort(bgr, bgrUndistorted, cameraMatrix, distortionCoefficients);
+        cv::undistort(bgr, bgrUndistortedFull, cameraMatrix, distortionCoefficients, optimalCameraMatrix);
 
-        std::cout << "Displaying and saving the RGB image" << std::endl;
+        std::cout << "Displaying and saving the BGR image" << std::endl;
 
-        displayRGB(rgb, "Distorted RGB image");
-        cv::imwrite("Distorted RGB image.jpg", rgb);
+        displayBGR(bgr, "Distorted BGR image");
+        cv::imwrite("Distorted RGB image.jpg", bgr);
 
-        std::cout << "Displaying and saving the undistorted RGB image" << std::endl;
+        std::cout << "Displaying and saving the undistorted BGR image" << std::endl;
 
-        displayRGB(rgbUndistorted, "Undistorted RGB image");
-        cv::imwrite("Undistorted RGB image.jpg", rgbUndistorted);
+        displayBGR(bgrUndistorted, "Undistorted BGR image");
+        cv::imwrite("Undistorted RGB image.jpg", bgrUndistorted);
 
-        std::cout << "Displaying and saving the Undistorted RGB image - full" << std::endl;
+        std::cout << "Displaying and saving the Undistorted BGR image - full" << std::endl;
 
-        displayRGB(rgbUndistortedFull, "Undistorted RGB image - full");
-        cv::imwrite("Undistorted RGB image - full.jpg", rgbUndistorted);
+        displayBGR(bgrUndistortedFull, "Undistorted BGR image - full");
+        cv::imwrite("Undistorted RGB image - full.jpg", bgrUndistorted);
     }
     catch(const std::exception &e)
     {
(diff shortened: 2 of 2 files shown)


knatten:
I think you as a team should decide on whether you want a different sample or not. What is the purpose of the  `CaptureUndistortRGB` sample? Is it to demonstrate how OpenCV can be used for undistortion? And what's the purpose of adding the 2D code? If it's just to demonstrate how to use the new 2D API, it's probably better if they're two separate samples. However, if the intention is to show that you can use both the 3D and the 2D api, it might make sense to have both in the same sample.

Then again, now that we have the 2D api, maybe we don't need to have a sample showing how to get 2D data from the 3D data at all? Or maybe it's still useful? I think you should discuss that. If we keep it, we should probably make a comment that it's just for demonstration, and that we recommend people to use the 2D api instead unless they need both 3D and 2D.

If you decide to keep both in one sample, I highly suggest taking user input to switch, or maybe just setting a `constexpr bool` or something. At least I recommend avoiding `#define`s. If we have code that's inside an undefined `#define`, it won't get compiled. So if that code goes out of date due to API changes, or changes elsewhere in the sample, we won't notice in CI.

Your comment about `waitKey()` in `64eae16` seems to refer to a commit that isn't part of this PR, I suggest you update the PR description to reflect what's currently in the PR, that makes it easier to review.

SatjaSivcev:
All the things you've mentioned we want to demonstrate: Capture2D, Capture2DFrom3D, UndirstortRGBWithOpenCV.
Regarding comments. I don't think we should recommend 2D over 2D from 3D. Each of these methods is better for some applications. That is up to a customer to decide. We need to demonstrate the possibilities. Of course, it is useful to comment when one might be used and when another.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,32 +27,64 @@ int main()
         std::cout << "Connecting to the camera" << std::endl;
         auto camera = zivid.connectCamera();
 
-        std::cout << "Configuring the camera settings" << std::endl;
-        camera << Zivid::Settings::Iris{ 21 } << Zivid::Settings::ExposureTime{ std::chrono::microseconds{ 20000 } }
-               << Zivid::Settings::Gain{ 1 } << Zivid::Settings::Brightness{ 1.0 };
+        std::cout << "Enter <a><Return> for 2D only, and just <Return> for 3D" << std::endl;
+        const auto command = getInput();
+        bool use2D = false;
+        if(command == "a" || command == "A")
```
knatten: Maybe using `2d` / `2D` instead of `a` / `A` would be more intuitive?

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,32 +27,64 @@ int main()
...
 
-        const auto pointCloud = frame.getPointCloud();
-        const auto rgb = pointCloudToRGB(pointCloud);
+            std::cout << "Running the visualizer. Blocking until the window closes" << std::endl;
+            vis.run();
+
+            std::cout << "Converting ZDF point cloud to OpenCV format" << std::endl;
+            const auto pointCloud = frame.getPointCloud();
+            rgb = pointCloudToRGB(pointCloud);
+        }
```
knatten: `main()` has gotten fairly long now. Maybe you could refactor out some code here? For instance you could have one function `getImage2D()` and one `getImage3D()` (each extracted from the respective if and else branches here).

A nice side effect is then that you don't have to remove `const` from `rgb`.  You could  do `const atuo rgb = use2d  ? getImage2D() : getImage3d();` or something like that.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,32 +27,64 @@ int main()
...
+            settings.set(Zivid::Settings2D::ExposureTime{ std::chrono::microseconds{ 10000 } });
+            settings.set(Zivid::Settings2D::Gain{ 1.0 });
+            settings.set(Zivid::Settings2D::Iris{ 35 });
+
+            std::cout << "Capture a 2D frame" << std::endl;
+            const auto frame = camera.capture2D(settings);
 
-        std::cout << "Capturing a frame" << std::endl;
-        const auto frame = camera.capture();
+            std::cout << "Get RGBA8 image from Frame2D" << std::endl;
```
knatten: You should be consistent with verb usage, the rest of the sample is using the continuous aspect ("Getting"  instead of "Get"). Same with "Capturing" vs "Capture" above.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,32 +27,64 @@ int main()
...
+        if(command == "a" || command == "A")
+        {
+            use2D = true;
+        }
+
+        cv::Mat rgb;
+
+        if(use2D)
+        {
+            std::cout << "2D only" << std::endl;
```
knatten: If you print the mode when using 2d, I suggest printing it when using 3d as well, for consistency.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,32 +27,64 @@ int main()
...
+
+            std::cout << "Configuring the camera settings" << std::endl;
+            auto settings = Zivid::Settings2D();
```
knatten: This is a matter of taste, but I personally find `Zivid::Settings2D settings` easier to read. What do you think?

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -87,6 +121,34 @@ int main()
...
+    cv::Mat rgb(static_cast<int>(image.height()), static_cast<int>(image.width()), CV_8UC3, cv::Scalar(0, 0, 0));
+
+    const auto height = image.height();
+    const auto width = image.width();
+
+    for(int i = 0; i < height; i++)
+    {
+        for(int j = 0; j < width; j++)
+        {
+            cv::Vec3b &color = rgb.at<cv::Vec3b>(i, j);
```
apartridge: You can convert without iterating over the whole image. Try this

        cv::Mat rgbaMat(height(), width(), CV_8UC4, const_cast<void *>(reinterpret_cast<const void *>(image.dataPtr())));
        cv::Mat rgbMat;
        cvtColor(rgbaMat, rgbMat, cv::COLOR_RGBA2RGB);

apartridge: This should be much faster

torbsorb: Except I would do:
cvtColor(rgbaMat, rgbMat, cv::COLOR_RGBA2BGR);

BGR instead of RGB (as OpenCV use BGR as default)

torbsorb: ...or modify the display function, and do the conversion there. All functions and variables indicate rgb, but `pointCloudToRGB` actually returns BGR. An alternative would be to rename all instances of rgb to bgr, where it's actually bgr.

knatten: Using the same format throughout the entire sample seems like a good idea. As well as renaming the functions and variables to reflect the correct rgb/bgr.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
torbsorb: Should use of auto be reserved to when the return type is somehow derived by a specific overload of a constructor? Are there any rules I should be aware of? Didn't immediately find clear guidance on when *not* to use auto.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -87,9 +80,79 @@ int main()
...
+std::string getInput()
+{
+    std::string command;
+    std::getline(std::cin, command);
+    return command;
+}
+
+cv::Mat imageToBGR(const Zivid::Image<Zivid::RGBA8> &image)
+{
+    cv::Mat rgbaMat(static_cast<int>(image.height()),
```
apartridge: why cast around height and not width?  You can probably remove the static_cast unless the CI complains about this.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -87,9 +80,79 @@ int main()
...
-cv::Mat pointCloudToRGB(const Zivid::PointCloud &pointCloud)
+cv::Mat getImage3D(Zivid::Camera &camera, Zivid::Application &zivid)
+{
+    std::cout << "3D mode" << std::endl;
+
+    std::cout << "Configuring the camera settings" << std::endl;
+    camera << Zivid::Settings::Iris{ 21 } << Zivid::Settings::ExposureTime{ std::chrono::microseconds{ 20000 } }
+           << Zivid::Settings::Gain{ 1 } << Zivid::Settings::Brightness{ 1.0 };
+
+    std::cout << "Capturing a frame" << std::endl;
```
SatjaSivcev: Should we specify in the comment that it is a 3D frame / 3D point cloud?

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -87,9 +80,79 @@ int main()
...
+    Zivid::Settings2D settings = Zivid::Settings2D();
+    settings.set(Zivid::Settings2D::ExposureTime{ std::chrono::microseconds{ 10000 } });
+    settings.set(Zivid::Settings2D::Gain{ 2.0 });
+    settings.set(Zivid::Settings2D::Iris{ 27 });
+    settings.set(Zivid::Settings2D::Brightness{ 1.0 });
+
+    std::cout << "Capturing a 2D frame" << std::endl;
+    const auto frame = camera.capture2D(settings);
+
+    std::cout << "Getting RGBA8 image from Frame2D" << std::endl;
```
SatjaSivcev: Why Frame2D and not 2D frame?

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,60 +31,47 @@ int main()
...
+        }
 
-        const auto pointCloud = frame.getPointCloud();
-        const auto rgb = pointCloudToRGB(pointCloud);
+        const auto bgr = (use2D ? getImage2D(camera) : getImage3D(camera, zivid));
 
-        std::cout << "Undistorting the RGB image" << std::endl;
+        std::cout << "Undistorting the BGR image" << std::endl;
 
+        auto intrinsics = camera.intrinsics();
```
SatjaSivcev: It seems you are not using this variable? Remove it if it is true.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
knatten: This is a topic of debate in the C++ community, and we don't have much about in our coding standards either. I'm fine with you using it here too, even though I wouldn't do it myself. I guess we should just agree on some guidelines.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
torbsorb: probably a replace all error

torbsorb:
Hi @SatjaSivcev, I noticed you merged master into this branch. I just locally rebased on master, and forced pushed that update. Should be the same output.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
torbsorb: I remember now.. I did it like this to check if the cast was necessary...

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -12,9 +12,15 @@ Undistort an RGB image from a ZDF point cloud using Zivid camera intrinsics.
...
+cv::Mat pointCloudToBGR(const Zivid::PointCloud &);
+cv::Mat imageToBGR(const Zivid::Image<Zivid::RGBA8> &);
 std::tuple<cv::Mat, cv::Mat> reformatCameraIntrinsics(const Zivid::CameraIntrinsics &);
-void displayRGB(const cv::Mat &, const std::string &);
+void displayBGR(const cv::Mat &, const std::string &);
+std::string getInput(void);
+
+cv::Mat getImage2D(Zivid::Camera &camera);
+
+cv::Mat getImage3D(Zivid::Camera &camera, Zivid::Application &zivid);
```
knatten: Is there a reason why you have a blank line between some of the forward declarations but not all?

torbsorb: no, used Visual Studio 2017 to refactor ;) didn't correct it

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -87,9 +79,79 @@ int main()
...
+std::string getInput()
+{
+    std::string command;
+    std::getline(std::cin, command);
+    return command;
+}
+
+cv::Mat imageToBGR(const Zivid::Image<Zivid::RGBA8> &image)
+{
+    cv::Mat rgbaMat(image.height(),
```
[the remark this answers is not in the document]
torbsorb: I guess I can since image is const, and I use const_cast? (I didn't originally write this, I wouldn't know to use const_cast)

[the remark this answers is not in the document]
knatten: Resolved irl

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,60 +29,46 @@ int main()
...
-        std::cout << "Displaying the point cloud" << std::endl;
-        vis.showMaximized();
-        vis.show(frame);
-        vis.resetToFit();
-
-        std::cout << "Running the visualizer. Blocking until the window closes" << std::endl;
-        vis.run();
-
-        std::cout << "Converting ZDF point cloud to OpenCV format" << std::endl;
+        std::cout << "Enter <2d><Return> for 2D only, and just <Return> for 3D" << std::endl;
```
SatjaSivcev: <2d><Return> is not the most intuitive description. I had to stop and remember that <Return> means hitting the 'Enter' key. It seems that it is traditionally 'return' on Mac PC and 'enter' on Windows.

It would've been more intuitive to me (this doesn't mean it would've be the best way) if it was like this: "Enter "2d" and press enter/return key for 2D only..."

Is the way you are describing options with "<>" a common approach? If it is, leave it (and educate me). If not, try to make it simpler (idiot-proof).

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -25,60 +29,46 @@ int main()
...
+        const auto command = getInput();
+        bool use2D = false;
+        if(command == "2d" || command == "2D")
```
SatjaSivcev: Instead of the blank for 3d I would add "3d" or "3D". It is more intuitive. Blank could work for one of them, or not, doesn't matter.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
torbsorb: I have no objection to what you propose. No reason not to be more explicit.

README.md
```diff
@@ -33,9 +33,9 @@ Check out our [tutorial on configuring and building these samples on Ubuntu](htt
 
 - [**ZividOpenCV**](https://github.com/zivid/cpp-extra-samples/tree/master/ZividOpenCV)
 	- [**ZDF2OpenCV**](https://github.com/zivid/cpp-extra-samples/blob/master/ZividOpenCV/ZDF2OpenCV/ZDF2OpenCV.cpp) - Import a ZDF point cloud and convert it to OpenCV format.
-	- [**CaptureUndistortRGB**](https://github.com/zivid/cpp-extra-samples/blob/master/ZividOpenCV/ZDF2OpenCV/CaptureUndistortRGB.cpp) - Undistort an RGB image from a ZDF point cloud using Zivid camera intrinsics.
+	- [**CaptureUndistortRGB**](https://github.com/zivid/cpp-extra-samples/tree/master/ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp) - Use Zivid camera intrinsics to undistort an RGB image. This example will prompt the user for whether to capture a 2D or a 3D image. In both instances it will operate on a 2D image. However, in the 3D case it will extract 2D image from a ZDF point cloud. The 2D variant is faster.
```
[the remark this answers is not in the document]
torbsorb: I believe tree is used for folder and blob for specific file(s). I'll change it.

ZividOpenCV/CaptureUndistortRGB/CaptureUndistortRGB.cpp
torbsorb: It would work with 3d or whatever really. Stating it explicitly would be untrue.

torbsorb:
Rebased after all comments were resolved. Please perform final review before merge.

PR #39, opened 2019-11-25
Enable warnings

knatten:
Summary: Set up the system for enabling warnings. Disable warnings by default. Enable warnings in CI, but not warnings as errors yet (have to fix the warnings first). See individual commit messages for details.

This PR also disables compiler extensions, and changes from Make to Ninja.


continuous-integration/linux/build.sh
```diff
@@ -2,12 +2,29 @@
...
+{
+    COMPILER=$1
+    BUILD_SUBDIR=$2
+    BUILD_DIR="$BUILD_ROOT_DIR/$BUILD_SUBDIR"
+
+    echo "-----------------------------------------------"
+    echo "            Building with $COMPILER"
+    echo "-----------------------------------------------"
+
+    cd $ROOT_DIR
```
trym-b: `|| exit $?`

continuous-integration/linux/build.sh
```diff
@@ -2,12 +2,29 @@
...
+        -DUSE_EIGEN3=OFF \
+        -DUSE_OPENCV=OFF \
+        -DWARNINGS=ON \
+        -DWARNINGS_AS_ERRORS=OFF \
+        ../.. || exit $?
+    cmake --build . || exit $?
+}
+
+build clang++-8 clang
+build g++-7 gcc
```
trym-b: `|| exit $?`

ZividOpenCV/ZDF2OpenCV/ZDF2OpenCV.cpp
```diff
@@ -130,9 +130,9 @@ int main()
         cv::applyColorMap(z, zJetColorMap, cv::COLORMAP_JET);
 
         // Setting nans to black
-        for(int i = 0; i < pointCloud.height(); i++)
+        for(size_t i = 0; i < pointCloud.height(); i++)
         {
-            for(int j = 0; j < pointCloud.width(); j++)
+            for(size_t j = 0; j < pointCloud.width(); j++)
```
trym-b: Why not `auto`?

knatten: Both works. Any reason to change it?

trym-b: In case whatever we are iterating over changes type?

continuous-integration/linux/build.sh
knatten: Fixed

continuous-integration/linux/build.sh
knatten: Fixed

ZividOpenCV/ZDF2OpenCV/ZDF2OpenCV.cpp
knatten: That could happen if Zivid breaks its api. I don't think it really matters, but have fixed it now.

continuous-integration/linux/build.sh
```diff
@@ -2,12 +2,29 @@
...
+    BUILD_SUBDIR=$2
+    BUILD_DIR="$BUILD_ROOT_DIR/$BUILD_SUBDIR"
+
+    echo "-----------------------------------------------"
+    echo "            Building with $COMPILER"
+    echo "-----------------------------------------------"
+
+    cd $ROOT_DIR || exit $?
+    mkdir -p $BUILD_DIR || exit $?
+    cd $BUILD_DIR || exit $?
```
trym-b: You can also do:
`mkdir --parents "$ROOT_DIR/$BUILD_DIR" || exit $?`
You should also use "" around variables containing paths, since there could be a space in one of them.

knatten: Good idea to use `--parents` instead of `-p`. Also removed the unnecessary `cd $ROOT_DIR`. Added quotes too.

knatten:
Ready for next round

knatten:
Pushed one more commit 8bf993a which uses default versions of clang and gcc instead of specific versions.

ZividOpenCV/ZDF2OpenCV/ZDF2OpenCV.cpp
knatten: Wait, `auto` won't work. `auto j = 0` deduces `int`. We want `size_t`. Reverting my fix.

cmake/CompilerOptions.cmake
```diff
@@ -58,5 +58,5 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /w")
     endif()
 else()
-    message(FATAL_ERROR "Unknown compiler: ${CMAKE_CXX_COMPILER_ID}")
+    message(WARNING "Unknown compiler, not able to set compiler options for ${CMAKE_CXX_COMPILER_ID}")
```
trym-b: Is `CMAKE_CXX_COMPILER_ID` a readable variable or do we need to add something like:
`message(WARNING "Unknown compiler, not able to set compiler options for compiler ID: ${CMAKE_CXX_COMPILER_ID}")`

knatten: What's a readable variable?

knatten:
This is on hold until I've been able to figure out why the `/experimental:external` doesn't work on @SatjaSivcev's computer. The rest of the stuff is ready.

cmake/CompilerOptions.cmake
trym-b: If the ID is `12345` or `a1fa335` I would say it is unreadable, but if it is `gcc--9.1.2` or `clang-10.1.2` I would say it is readable.

knatten: Ah, now I understand what you mean. Good point! It's "Clang", "GNU" and "MSVC" for those three compilers. I assume others are similarly readable.

knatten:
@trym-b any further comments before I squash, rebase and merge?

knatten:
Just talked to @trym-b irl, he has no further objections.

PR #40, opened 2019-11-27
Upgrade to Zivid 1.7

SatjaSivcev:


SatjaSivcev:
I knew I forgot to do something. Done, ready for the next round.

knatten:
A tip when changing things like these is to do `git grep 1.6.0` to quickly search the repo for the thing you're changing. It's saved me a few times in the past! :)

PR #45, opened 2019-11-29
Modify to work with official Eigen

SatjaSivcev:


Downsample/Downsample.cpp
```diff
@@ -138,17 +138,26 @@ Eigen::MatrixXf gridSum(const Eigen::MatrixXf &r, int downsamplingFactor)
 
 Eigen::MatrixXf lineSum(const Eigen::MatrixXf &matrix, int downsamplingFactor)
 {
+    Eigen::Map<Eigen::MatrixXf> reshapedMatrixMap(const_cast<float *>(matrix.data()),
```
knatten: This `const_cast` is safe because we don't actually alter the data, is that it?

Downsample/Downsample.cpp
```diff
@@ -138,17 +138,26 @@ Eigen::MatrixXf gridSum(const Eigen::MatrixXf &r, int downsamplingFactor)
...
 
     reshapedMatrix = reshapedMatrix.unaryExpr(nan_to_zero_functor);
 
-    return reshapedMatrix.colwise()
-        .sum()
-        .reshaped((int)matrix.rows() / downsamplingFactor, (int)matrix.cols())
-        .transpose();
+    Eigen::MatrixXf reshapedMatrixColwiseSum = reshapedMatrix.colwise().sum();
+
+    Eigen::Map<Eigen::MatrixXf> reshapedMatrixMap2(reshapedMatrixColwiseSum.data(),
```
knatten: Can you find more descriptive names for `reshapedMatrixMap2` and `reshapedMatrix2`?

Downsample/Downsample.cpp
```diff
@@ -138,17 +138,26 @@ Eigen::MatrixXf gridSum(const Eigen::MatrixXf &r, int downsamplingFactor)
...
+                                                   (int)matrix.rows() / downsamplingFactor,
+                                                   (int)matrix.cols());
+
+    Eigen::MatrixXf reshapedMatrix2 =
+        Eigen::Map<Eigen::MatrixXf>(reshapedMatrixMap2.data(), reshapedMatrixMap2.rows(), reshapedMatrixMap2.cols());
+
+    return reshapedMatrix2.transpose();
```
knatten: I don't know Eigen, how does `transpose` work? The docs say it returns "an expression of the transpose of *this.", whatever that means. Sounds like it's some sort of lazy thing which refers back to `reshapedMatrix2` which goes out of scope when the function exits? Is that safe?

Downsample/Downsample.cpp
```diff
@@ -138,17 +138,26 @@ Eigen::MatrixXf gridSum(const Eigen::MatrixXf &r, int downsamplingFactor)
...
+    Eigen::Map<Eigen::MatrixXf> reshapedMatrixMap(const_cast<float *>(matrix.data()),
+                                                  downsamplingFactor,
+                                                  (int)matrix.rows() * (int)matrix.cols() / downsamplingFactor);
     Eigen::MatrixXf reshapedMatrix =
-        matrix.reshaped(downsamplingFactor, (int)matrix.rows() * (int)matrix.cols() / downsamplingFactor);
+        Eigen::Map<Eigen::MatrixXf>(reshapedMatrixMap.data(), reshapedMatrixMap.rows(), reshapedMatrixMap.cols());
 
     std::function<float(float)> nan_to_zero_functor = nanToZero;
 
     reshapedMatrix = reshapedMatrix.unaryExpr(nan_to_zero_functor);
```
knatten: Can we use a new variable here, instead of assigning to the old one?

knatten: Do we actually modify some of the data in `reshapedMatrix` which we casted away `const` for here?

knatten:
@runenordmo can you help reviewing this?

Downsample/Downsample.cpp
SatjaSivcev: Yes, the input matrix to this funciton is not changed. A new one is created from it.

knatten: Does `matrix.data()` actually get copied? Or is that data pointer just being wrapped in a new data structure, and the new structure keeps pointing to the old data?

SatjaSivcev: I don't know. How to check that?

Downsample/Downsample.cpp
SatjaSivcev: flattenedMatrix and reshapedMatrix

Downsample/Downsample.cpp
SatjaSivcev: Eigen::MatrixXf reshapedMatrix =
        Eigen::Map<Eigen::MatrixXf>(reshapedMatrixMap.data(), reshapedMatrixMap.rows(), reshapedMatrixMap.cols())
            .transpose();

    return reshapedMatrix;

Is this better?

Downsample/Downsample.cpp
SatjaSivcev: Yes. It is really difficult to come up with names for variables that are just part of the process. Is there any convention on this? I am inventing names to explain the operation.

SatjaSivcev: flattenedMatrixNansRemoved is the best I could think of.

SatjaSivcev: Yes we replace NaNs with Zeroes to be able to sum data.

Downsample/Downsample.cpp
SatjaSivcev: done

Downsample/Downsample.cpp
SatjaSivcev: done

SatjaSivcev:
Ready for the next round. I haven't resolved all the comments.

SatjaSivcev:
Why can't I add Rune as a reviewer?

Downsample/Downsample.cpp
knatten: You have to read the Eigen docs to figure that out.

Downsample/Downsample.cpp
knatten: I don't know. Someone who knows Eigen has to figure this out.

Downsample/Downsample.cpp
knatten: So the underlying data on which we replace NaNs with Zeros is actually the data which we previously cast away `const` from? If that's the case, we have to change this. We can't take a `const` parameter to a function, cast away `const` from it, and then mutate the data.

knatten:
I've added him to the organisation now, so you can add him.

Downsample/Downsample.cpp
runenordmo: I think the map is only mapping a existing array, specifying the alignment and the stride. Normally, modifying the `Map` will also modify the `matrix`. But I think we should do the rest of the review live, as @knatten suggested.

SatjaSivcev:
Ready for the next round.

Downsample/Downsample.cpp
SatjaSivcev: .transpose has been moved out of the lineSum()

runenordmo:
Looks good to me too 👍

SatjaSivcev:
I've changed the commit message. Is it too long?

Ready for the next round.

knatten:
I think it's very good! It tells me exactly what I need to know when I come back in two months and don't remember why this change was made.

The only thing I would change is adding a short header, and wrap it properly. Your editor can probably be set up to do this automatically.

[How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) is a great introduction to good commit messages, and there's also some info in the internal "Development Workflow" wiki.

Here's my suggestion:
> Avoid depending on unreleased Eigen in Downsample
>
> Modify the sample to make it compatible with the latest stable Eigen
> release (3.3.7) instead of the latest unstable source code from the
> development branch (3.3.90)

SatjaSivcev:
Done. Thanks.

SatjaSivcev:
I've just modified the Readme. Please review.

knatten:
LGTM! :)

Issue #43, opened 2019-11-29
Add all dependencies in CI


knatten:
Currently we only have PCL, not Eigen nor OpenCV. And when adding 16.04, we'll not even have PCL on that one.

Consider using Conan like we do in zivid-python. Think about the experience for people who just want to clone the repo and check out the samples, need to still be easy to get started.

Here's a summary of which versions we currently use (according to `README.md on tip of master`), and which are available in Ubuntu 16 / 18 / 20:

| Dependency | We use | Ubuntu 16.04 | Ubuntu 18.04  | Ubuntu 20.04 |
|---|---|---|---|---|
| Eigen | 3.3.7 |3.3~beta1-2  | 3.3.4 | 3.3.7 |
| OpenCV | 4.0.1 | 2.4.9 | 3.2.0 | 4.2.0 |
| PCL | ???| 1.7.2 | 1.8.1 | 1.10.0 |

nedrebo:
Can probably do the canon stuff only if USE_PCL or USE_EIGEN is ON.

CI of course sets everything to ON.

knatten:
Yeah that would be good.

The main work for this issue I think is to decide whether to ask users to use Conan as well. We currently have [a quite long KB article](https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/61472793/Configure+C+Extra+Samples+with+CMake+and+build+them+using+Visual+Studio+in+Windows) for how to install everything correctly on Windows, and no instructions for Linux. We need to figure out if it's simpler to instruct people to install conan instead, and allow dependencies to be installed automatically.

If we do that, it should still be possible to clone this repo and build the samples which have no dependencies, without bothering about conan. (Like it is today.)

Also need to figure out if we should allow people to override these settings and use an installation of Eigen not from conan etc.

nedrebo:
> Also need to figure out if we should allow people to override these settings and use an installation of Eigen not from conan etc.

I would say no to this one. Sounds like advanced usage, they figure it out anyway. Keep it as simple as possible, optimize for the casual user.

knatten:
Blocked by #44

knatten:
When this is fixed, search the code base for `#43` and remove any workarounds which have been introduced in the mean time.

SatjaSivcev:
Ubuntu 20.04  LTS is out, with these versions:
https://www.ubuntuupdates.org/package/core/focal/universe/base/opencv
[...]

PR #50, opened 2019-12-02
Add CI for Ubuntu 16.04

knatten:
There's unfortunately a bug in Zivid which causes issues on Clang 3. An
internal issue has been logged for this.

### DIFF
--- a/continuous-integration/linux/build.sh
+++ b/continuous-integration/linux/build.sh
@@ -4,6 +4,20 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR=$(realpath "$SCRIPT_DIR/../..")
 BUILD_ROOT_DIR="$ROOT_DIR/build"
 
+#There's a C++11 compatibility bug in the Zivid API which makes it fail on
+#older versions of Clang, such as the default Clang 3.8.2 on Ubuntu 16.04.
+#An internal Zivid issue to fix this has been logged. When that issue has
+#been fixed, this entire option should be removed.
+if [[ "$1" == "--skip-clang" ]]; then
+    SKIP_CLANG=true
+fi
+
+#TODO 16.04 default pcl is too old. See issue #43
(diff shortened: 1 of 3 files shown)


continuous-integration/linux/build.sh
```diff
@@ -4,6 +4,12 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR=$(realpath "$SCRIPT_DIR/../..")
 BUILD_ROOT_DIR="$ROOT_DIR/build"
 
+#TODO 16.04 default pcl is too old. See issue #43
+source /etc/os-release || exit $?
+if [[ "$VERSION_ID" == "16.04" ]]; then
+    OS_SPECIFIC_OPTIONS="-DUSE_PCL=OFF"
```
nedrebo: Looking at `azure-pipelines.yml` it looks like 16.04 is first class citizen.

I wonder if `build.sh` should be kept platform independent, then we have two options:
 1. This and the `if` below should be handled as cli options to build.sh so that we have full transparency of the quirks in `azure-pipelines.yml`.
 2. Fix `ubuntu-16.04/setup.sh` so that build.sh works. Then we also document how to fully set up a working environment.
 3. Combinations of the above

I prefer option (2.). Do all the fixes/quirks in `setup.sh` leave `build.sh` unchanged and keep `azure-pipelines.yml` clean.

knatten: 16.04 should absolutely be a first class citizen. This `if` goes away when I do #43, which is very high pri.

After #43 is fixed, "fix `setup.sh` so that `build.sh` works" would mean installing a non system-default version of clang. Since we claim to support 16.04, I think it is important to test on the default clang on 16.04. We have an internal issue in Zivid to fix Clang 3 support, so the second `if` will hopefully also go away in a few months. In the mean time, I think option 1 is best.

knatten: @nedrebo I've pushed a suggestion now, what do you think?

PR #52, opened 2019-12-03
Modify code to fix warnings

SatjaSivcev:


Downsample/Downsample.cpp
```diff
@@ -106,11 +106,13 @@ Zivid::PointCloud downsample(const Zivid::PointCloud &pointCloud, int downsampli
 
     Zivid::PointCloud pointCloudDownsampled(redDownsampled.rows(), redDownsampled.cols());
 
-    for(int i = 0; i < redDownsampled.rows(); i++)
+    for(size_t i = 0; i < static_cast<size_t>(redDownsampled.rows()); i++)
```
nedrebo: Remove the `static_cast`. I do not think there should be any reason to do static_casts or any other casts for that case in this code base at all.

A `static_cast` like this does not change the behavior of your program, the only thing that happens is that you suppress a warning. If you do not care about the warning you should disable it, if you care you need to fix the underlying problem.

Typically when one crank up the warning level in a project developers start to code things that are worse because what they used to do is no longer allowed. And, since they don't get the underlying idea of the warning and they (unintentionally)  hack around it.

I recommend that you refuse to do a change just because a warning tells you to, you need to personally agree that adding the `static_cast<size_t>(` is an improvement. I would argue the only thing this change does is to reduce readability.

I think all in CS team should only do changes they approve of. When you and the warnings are misaligned you need to ask someone what is the reason for having this warning, and what is the correct fix.

cmake/CompilerOptions.cmake
```diff
@@ -50,12 +50,13 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
...
             4702 # Got unreachable warnings from external, even though they are generallly ignored by /experimental:external
-            4710 # If the compiler decides to not inline a function, that's their decision.
-            4711 # If the compiler decides to inline a function, that's their decision.
+            4710 # If the compiler decides to not inline a function, that's their decision
+            4711 # If the compiler decides to inline a function, that's their decision
             4571 # Just a non-interesting informational warning about msvc changing behaviour in 7.1
             4267 # Conversion: Happens a lot in these samples, would complicate them too much to handle manually
             4365 # Conversion: Happens a lot in these samples, would complicate them too much to handle manually
             4388 # Signed/unsigned comparison: Happens a lot in these samples, would complicate them too much to handle manually
+            5045 # This warning is purely informational
```
nedrebo: I think it would be nice if the warning error text was included (stripping away concrete type names if any). This is hard to understand without context. Like this:

```
5045 # Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
         # This warning is purely informational
```

This goes for all the warnings here

continuous-integration/linux/build.sh
```diff
@@ -34,7 +34,7 @@ function build()
         -DUSE_EIGEN3=OFF \
         -DUSE_OPENCV=OFF \
         -DWARNINGS=ON \
-        -DWARNINGS_AS_ERRORS=OFF \
+        -DWARNINGS_AS_ERRORS=ON \
```
nedrebo: NIce

Downsample/Downsample.cpp
knatten: The good thing with requiring `static_cast` is that you're forced to make a decision whether your narrowing is safe or not. But I agree they make the code uglier, and these warnings might be too strict for this repo. I'm fine with disabling narrowing warnings.

SatjaSivcev: It is int again.

cmake/CompilerOptions.cmake
SatjaSivcev: done

SatjaSivcev:
Ready fo the next round.

SatjaSivcev:
Done. Enable warnings as errors is the last commit.

PR #54, opened 2019-12-06
Opencv dependencies

knatten:
Make OpenCV components non-optional, and add a missing OpenCV dependency.


knatten:
Rebased on master and force pushed now

PR #56, opened 2019-12-10
Fix warnings

SatjaSivcev:


HandEyeCalibration/PoseConversions/PoseConversions.cpp
```diff
@@ -23,10 +23,9 @@ enum class RotationConvention
     ZYX_Intrinsic,
     XYZ_Extrinsic,
     XYZ_Intrinsic,
-    ZYX_Extrinsic,
-    NOF_ROT
+    ZYX_Extrinsic
 };
-constexpr size_t nofRotationConventions = static_cast<size_t>(RotationConvention::NOF_ROT);
+constexpr size_t nofRotationConventions = static_cast<size_t>(RotationConvention::ZYX_Extrinsic);
```
knatten: Did you test this? Won't you now skip `ZYX_Extrinsic` when iterating?

cmake/CompilerOptions.cmake
```diff
@@ -61,6 +61,7 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
                  # This warning is purely informational
             4244 # Narrowing conversions: Too strict and noisy for this code base
             4242 # Narrowing conversions: Too strict and noisy for this code base
+            4820 # The type and order of elements caused the compiler to add padding to the end of a struct.
```
knatten: It would be good if the commit message included info about what parts of the code triggered this warning.

HandEyeCalibration/PoseConversions/PoseConversions.cpp
```diff
@@ -256,6 +257,7 @@ Eigen::Matrix3d rollPitchYawToRotationMatrix(const Eigen::Array3d &rollPitchYaw,
                     * Eigen::AngleAxisd(rollPitchYaw[0], Eigen::Vector3d::UnitX()))
                 .matrix();
             break;
```
nedrebo: Unreachable code. Clang has a warning for this, I expect it to cover this. You should enable it, if it is already I am surprised it does not fire on this. 

The break after return will never be hit, you should delete it.

knatten: This sample is not tested in CI due to #43, so Clang hasn't had the chance to complain yet.

HandEyeCalibration/PoseConversions/PoseConversions.cpp
SatjaSivcev: I've added + 1 to fix this. Not sure if there is a better way, but it works as it should now.
It seems there is no way to get the size of elements in an enum, except like we had with NOF_ROT which was used as a 'Count' from here:
https://stackoverflow.com/questions/14989274/is-it-possible-to-determine-the-number-of-elements-of-a-c-enum-class
The point is that we want this constexpr to be equal to the size of the enum (even if the enum is expanded). Any suggestions for a better method?

cmake/CompilerOptions.cmake
SatjaSivcev: This is the code:

`struct RollPitchYaw
{
    RotationConvention convention;
    Eigen::Array3d rollPitchYaw;
};

struct Representations
{
    Eigen::AngleAxisd axisAngle;
    Eigen::Vector3d rotationVector;
    Eigen::Quaterniond quaternion;
    std::array<RollPitchYaw, nofRotationConventions> rotations;
};`

This is the warning:

warning C4820: 'RollPitchYaw': '4' bytes padding added after data member 'RollPitchYaw::convention'
warning C4820: 'Representations': '8' bytes padding added after data member 'Representations::rotationVector'

How does the line below sound?

The type and order of elements caused the compiler to add padding to the end of a struct; specifically: RollPitchYaw::convention and Representations::rotationVector

If we add more code that would trigger this warning we would not catch it in the future because we have already disabled the warning. Then the comments with info about what parts of the code triggered this warning would not be complete right? Is this then the best practice?

HandEyeCalibration/PoseConversions/PoseConversions.cpp
SatjaSivcev: Removed

SatjaSivcev:
Ready for the next round.

HandEyeCalibration/PoseConversions/PoseConversions.cpp
knatten: Tip: Since most samples aren't built in CI yet, it's a good idea to enable warnings when building locally.

cmake/CompilerOptions.cmake
knatten: The good thing about comments in git commits is that they are tied to a specific point in time. When I read a git commit comment, I know that it's a comment about the state of the code right then. I do not expect it to be valid in the future.

(This is opposed to comments in the source code itself, which need to be kept up to date as the code changes.)

Git commit comments are very useful when someone in the future wonders why a certain thing was done in the past. Say I want to know why this warning was ignored. Then it's very easy to find the commit which ignored it, and I should then be able to read an explanation in the commit message.

I think your suggested commit message there is good.

HandEyeCalibration/PoseConversions/PoseConversions.cpp
knatten: There's indeed no way to get the count of elements in an enum. I think the `NOF_ROT` trick is ok, it's been so long since I last looked at this PR that I don't remember if we discussed whether/why to remove it. Do you?

SatjaSivcev: Because we get a warning that it is unused.

knatten:
Looks good. Please clean up the history into two commits (one for the `NOF_ROT` thing, and one for 4820) and rebase on master.

Did you check whether this introduces any new warnings on Clang/GCC?

SatjaSivcev:
I've rebased.
No, how can I check whether this introduces any new warnings on Clang/GCC?

knatten:
> No, how can I check whether this introduces any new warnings on Clang/GCC?

Build it locally on Linux with warnings enabled

SatjaSivcev:
Ok, I will do that. Is there something blocking us to do that on the CI?

knatten:
#43 again. Samples which depend on Eigen or OpenCV are not built in CI.

SatjaSivcev:
Clang:

/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/HandEyeCalibration/UtilizeEyeInHandCalibration/UtilizeEyeInHandCalibration.cpp:12:17: warning: no previous prototype for function
      'cvToEigen' [-Wmissing-prototypes]
Eigen::MatrixXd cvToEigen(const cv::Mat &cvMat)
                ^
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/HandEyeCalibration/UtilizeEyeInHandCalibration/UtilizeEyeInHandCalibration.cpp:32:9: warning: no previous prototype for function
      'readTransform' [-Wmissing-prototypes]
cv::Mat readTransform(const std::string &file_name)
        ^
[ 91%] Linking CXX executable CaptureUndistortRGB
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/Downsample/Downsample.cpp:113:48: warning: implicit conversion loses integer precision: 'Eigen::DenseCoeffsBase<Eigen::Matrix<int, -1, -1,
      0, -1, -1>, 1>::Scalar' (aka 'int') to 'unsigned char' [-Wconversion]
            pointCloudDownsampled(i, j).setRgb(redDownsampled(i, j), greenDownsampled(i, j), blueDownsampled(i, j));
                                        ~~~~~~ ^~~~~~~~~~~~~~~~~~~~
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/Downsample/Downsample.cpp:113:70: warning: implicit conversion loses integer precision: 'Eigen::DenseCoeffsBase<Eigen::Matrix<int, -1, -1,
      0, -1, -1>, 1>::Scalar' (aka 'int') to 'unsigned char' [-Wconversion]
            pointCloudDownsampled(i, j).setRgb(redDownsampled(i, j), greenDownsampled(i, j), blueDownsampled(i, j));
                                        ~~~~~~                       ^~~~~~~~~~~~~~~~~~~~~~
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/Downsample/Downsample.cpp:113:94: warning: implicit conversion loses integer precision: 'Eigen::DenseCoeffsBase<Eigen::Matrix<int, -1, -1,
      0, -1, -1>, 1>::Scalar' (aka 'int') to 'unsigned char' [-Wconversion]
            pointCloudDownsampled(i, j).setRgb(redDownsampled(i, j), greenDownsampled(i, j), blueDownsampled(i, j));
                                        ~~~~~~                                               ^~~~~~~~~~~~~~~~~~~~~
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/HandEyeCalibration/PoseConversions/PoseConversions.cpp:34:20: warning: padding struct 'RollPitchYaw' with 4 bytes to align 'rollPitchYaw'
      [-Wpadded]
    Eigen::Array3d rollPitchYaw;
                   ^
[ 91%] Built target CaptureUndistortRGB
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/Downsample/Downsample.cpp:143:52: warning: use of old-style cast [-Wold-style-cast]
                                                   (int)matrix.rows() * (int)matrix.cols() / downsamplingFactor);
                                                   ^    ~~~~~~~~~~~~~
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/Downsample/Downsample.cpp:143:73: warning: use of old-style cast [-Wold-style-cast]
                                                   (int)matrix.rows() * (int)matrix.cols() / downsamplingFactor);
                                                                        ^    ~~~~~~~~~~~~~
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/HandEyeCalibration/PoseConversions/PoseConversions.cpp:41:24: warning: padding struct 'Representations' with 8 bytes to align 'quaternion'
      [-Wpadded]
    Eigen::Quaterniond quaternion;
                       ^
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/Downsample/Downsample.cpp:151:51: warning: use of old-style cast [-Wold-style-cast]
                                                  (int)matrix.rows() / downsamplingFactor,
                                                  ^    ~~~~~~~~~~~~~
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/Downsample/Downsample.cpp:152:51: warning: use of old-style cast [-Wold-style-cast]
                                                  (int)matrix.cols());
                                                  ^    ~~~~~~~~~~~~~

GCC:

/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/HandEyeCalibration/PoseConversions/PoseConversions.cpp: In function ‘Representations zividToRobot(const Affine3d&)’:
/home/csu/samples/cpp-extra-samples/source/Applications/Advanced/HandEyeCalibration/PoseConversions/PoseConversions.cpp:158:22: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     for(int i = 0; i < nofRotationConventions; i++)
                    ~~^~~~~~~~~~~~~~~~~~~~~~~~

SatjaSivcev:
@knatten what is your advice in general on the warning: no previous prototype for function - implement i by always declaring functions and make this our practice? Or disable it? For consistency I would implement it every time.

knatten:
> what is your advice in general on the warning: no previous prototype for function - implement i by always declaring functions and make this our practice? Or disable it? For consistency I would implement it every time.

Usually one puts these functions inside an anonymous namespace. So you do
```cpp
namespace{
    void foo(){}
}
int main()
{
    foo();
}
```

But there is also a nice advantage to what you did here: Since you added forward declarations and moved the helper functions to the bottom of the file, `main()` is now the first thing you see when you start reading the sample. I think that is nice, and makes the sample easier to read. I suggest keeping your fix.

source/cmake/CompilerOptions.cmake
```diff
@@ -21,6 +21,8 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
             sign-conversion                 # Happens a lot in these samples, would complicate them too much to handle manually
             sign-compare                    # Happens a lot in these samples, would complicate them too much to handle manually
             shorten-64-to-32                # Narrowing conversions: Too strict and noisy for this code base
+            padded                          # The type and order of elements caused the compiler to add padding to the end of a struct; specifically: RollPitchYaw::convention and Representations::rotationVector
+            conversion                      # Implicit conversion loses integer precision (signed to unsigned): Expected to happen a lot in these samples, would complicate them too much to handle manually; specifically: pointCloudDownsampled(i, j).setRgb(redDownsampled(i, j), greenDownsampled(i, j), blueDownsampled(i, j));
```
knatten: I think the information about which code triggered these warnings belong in the commit message rather than in the code. Comments in a commit message are about the state of the code as that commit was made, whereas comments in the code are expected to keep being up to date with the current state of the code. 

So here in the `.cmake` file, I'd just have
```cmake
padded # The type and order of elements caused the compiler to add padding to the end of a struct
conversion # Implicit conversion loses integer precision (signed to unsigned). Expected to happen a lot in these samples, would complicate them too much to handle manually

# (...)

4820 # The type and order of elements caused the compiler to add padding to the end of a struct
```

And then move the rest of the info to the commit message,

knatten:
Just tested this locally (with all dependencies enabled), now I get no warnings from neither clang 9 nor gcc 7! :trophy:

SatjaSivcev:
> 
> 
> Just tested this locally (with all dependencies enabled), now I get no warnings from neither clang 9 nor gcc 7! 🏆

Oh great, I was setting up a virtual machine to do the same.

knatten:
> Oh great, I was setting up a virtual machine to do the same.

That should come in handy in the future anyway! :)

source/cmake/CompilerOptions.cmake
SatjaSivcev: done

SatjaSivcev:
This should be it. Ready for the next round.
I was trying to rename the branch but I wasn't sure how to and didn't want to risk breaking something. It would be better if it didn't include 'in-pose-conversions' in the name since the warnings are from multiple cpp files.

knatten:
> I was trying to rename the branch but I wasn't sure how to and didn't want to risk breaking something. It would be better if it didn't include 'in-pose-conversions' in the name since the warnings are from multiple cpp files.

You'd have to make a new PR in that case. I don't think it's important enough to bother with.

PR #59, opened 2020-01-24
Add save frame to PLY

SatjaSivcev:


SatjaSivcev:
I am not sure if this should be merged as it would then deviate from the sample code that comes with the Zivid installer.

PR #60, opened 2020-01-27
Expand capture hdr

SatjaSivcev:


Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        auto hdrFrame = Zivid::HDR::capture(camera, settingsVector);
+
+        std::cout << "Acquiring HDR source images and combining them into HDR frame (slower)" << std::endl;
+        std::vector<Zivid::Frame> frames;
+
+        for(size_t i = 0; i < 3; ++i)
+        {
+            camera << Zivid::Settings::Iris{ iris[i] }
```
knatten: Why did you move all these settings inside the loop? I think it was better before when all the non-changing settings were applied once.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,100 @@
...
+        for(size_t i = 0; i < 3; ++i)
+        {
+            settingsVector.push_back(
+                Zivid::Settings{}
+                    .set(Zivid::Settings::Iris{ iris[i] })
+                    .set(Zivid::Settings::ExposureTime{ std::chrono::microseconds{ exposureTime[i] } })
+                    .set(Zivid::Settings::Brightness{ 1 })
+                    .set(Zivid::Settings::Gain{ 1 })
+                    .set(Zivid::Settings::Bidirectional{ false })
+                    .set(Zivid::Settings::Filters::Contrast::Enabled::yes)
```
knatten: If the goal of this example is to demonstrate HDR captures, I don't think we need to set all these settings. It just complicates the sample. I know it was like that before too, but maybe we could remove most of them for both the old and the new code.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,100 @@
...
+                   << Zivid::Settings::Filters::Reflection::Enabled::yes
+                   << Zivid::Settings::Filters::Saturated::Enabled::yes << Zivid::Settings::BlueBalance{ 1.081 }
+                   << Zivid::Settings::RedBalance{ 1.709 };
+
+            frames.emplace_back(camera.capture());
+            std::cout << "Frame " << i << " " << camera.settings() << std::endl;
+        }
+
+        std::cout << "Creating the HDR frame" << std::endl;
+        hdrFrame = Zivid::HDR::combineFrames(begin(frames), end(frames));
```
knatten: I don't like re-using variables. I think it would be nicer if you used two separate variables here, and then both could also be `const`. Makes it easier to reason about the code when you know stuff won't change, you worry less about the scope, side effects etc.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        std::cout << "Saving the frames" << std::endl;
```
knatten: I was a bit surprised that you save and display only the frame(s) from the last of the two hdr methods.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
SatjaSivcev: That is the part of what we want to show in this sample, i.e. how to set all possible settings. We want to allow the users to just copy-paste what they need from the settings.

We have removed the sample CaptureHDRCompleteSettings and I am trying to add that code in this sample.

Perhaps the description should be modified instead to specify why we are configuring complete settings.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
SatjaSivcev: Done

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
SatjaSivcev: We actually don't need to display anything here. That is covered elsewhere. I modified to save all frames now.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
knatten: Aren't "How do I make an HDR capture" and "What's the code for setting all possible settings" two different questions which would better be explained by two different samples?

> We have removed the sample CaptureHDRCompleteSettings

When was that removed?

SatjaSivcev: 'How do I make an HDR capture' is answered in [CaptureHDR](https://github.com/zivid/cpp-extra-samples/blob/master/Camera/Basic/CaptureHDR/CaptureHDR.cpp) (this one should only have Zivid::HDR::capture in my opinion because that is the fast method but that is a different story)

This example aims to give a bit more than just how to capture an HDR, it aims to demonstrate a) two different ways of taking an HDR, b) changing settings before and in a loop, c) configuring all settings.

We started having too many samples so it was decided that CaptureHDRCompleteSettings was to be coved by CaptureHDRLoop, and it was removed in this [commit](https://github.com/zivid/cpp-extra-samples/commit/84990f969440d0b1a1b516391ed78a19c2a56000).

Maybe we should sit and decide how we deal with a) b) c). @torbsorb @knatten?

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
SatjaSivcev: Done

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
knatten: I would probably have done it differently, but if you want to do all of this in this sample, that's up to you. You have a better overview of what needs doing and what's done elsewhere than I do.

Regarding removal of `CaptureHDRCompleteSettings`: I think it's important to note in commit messages if you remove stuff. "Reorganize samples" sounds like samples were just moved around. If you deleted stuff or consolidated some samples in that commit, it would have been good if it was mentioned in the message. 

Sitting down together and looking at this sounds like a good idea.

torbsorb:
We agreed to remove CaptureHDRLoop, right? Then create one example which reads settings from file, and one which sets all settings in code?

SatjaSivcev:
We have agreed to create one example which reads settings from file, and one which sets all settings in code. We have also agreed to remove CaptureHDRLoop as it was not consistent between other programming language code samples. My mind is not clear if we want to remove CaptureHDRLoop for other programming languages. That is why I started modifying this one. We can decide tomorrow.

SatjaSivcev:
Ready for the next round.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,60 @@
...
+                                                          { 90000, 40000, 10000 } };
+
+        for(size_t i = 0; i < 3; ++i)
+        {
+            std::vector<Zivid::Settings> settingsHDR;
+            for(size_t j = 0; j < 3; ++j)
+            {
+                settingsHDR.push_back(
+                    settings.set(Zivid::Settings::Iris{ iris[j] })
+                        .set(Zivid::Settings::ExposureTime{ std::chrono::microseconds{ exposureTimeFrames[i][j] } }));
```
torbsorb: Shouldn't we use emplace_back? Do you know whether it's doing a copy, or move? I tried to read the push_back docs. Just worried we'll have three identical frames in each HDR.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,60 @@
...
+            }
+            settingsHDRVector.push_back(settingsHDR);
```
torbsorb: Here push_back should be ok, although I don't see why a vector is needed here. We're not referencing this vector of frame settings outside of the loop.

Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp
```diff
@@ -0,0 +1,73 @@
...
+            .set(Zivid::Settings::Filters::Gaussian::Enabled::yes)
+            .set(Zivid::Settings::Filters::Gaussian::Sigma{ 1.5 })
+            .set(Zivid::Settings::Filters::Outlier::Enabled::yes)
+            .set(Zivid::Settings::Filters::Outlier::Threshold{ 5 })
+            .set(Zivid::Settings::Filters::Reflection::Enabled::yes)
+            .set(Zivid::Settings::Filters::Saturated::Enabled::yes)
+            .set(Zivid::Settings::BlueBalance{ 1.081 })
+            .set(Zivid::Settings::RedBalance{ 1.709 });
+
+        std::cout << "Configuring settings different for all HDR frames" << std::endl;
```
torbsorb: either "differently", or "Configuring different settings for all HDR frames". But, shouldn't it also say ".. for all frames in the HDR merge". Just to clarify that it's one HDR that consists of multiple frame, and not multiple HDRs... It should be clear... but it might not be.

SatjaSivcev: I agree it should be clear, not easy though. This is my current proposal:

Configuring settings same for all frames in the HDR frame
Configuring settings differently for all frames in the HDR frame

I wanted to get rid of 'merge'. Both single captures and HDRs are frames so I think above makes sense?

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
SatjaSivcev: Is this what you would have done?
`const` auto hdrFrame = Zivid::HDR::capture(camera, `settingsHDR);`
I thought it was nice for educational purposes to show how to create a vector of multiple HDR settings. Perhaps it is not needed as you say.
To be honest, I forgot what we agreed the purpose of this sample is to be. I remember that we've decided to keep CaptureHDRLoop and CaptureHDRCompleteSettings but I don't remember what we've agreed to emphesize in each of them.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
SatjaSivcev: For `void push_back (const value_type& val);` it says that val is: value to be copied (or moved) to the new element.
For `void emplace_back (Args&&... args);` it says that args are: Arguments forwarded to construct the new element. This new element is constructed in place using args as the arguments for its constructor.

I think it doesn't matter which one we use, and is, in fact, the same in our case, as `settings` object already is created. It seems that it would matter only if the object didn't already exist and we wanted to pass just the arguments for creating settings object. Then we would have to use emplace_back.

@knatten can you help clarify this?

Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp
torbsorb: Configuring settings differently for all frames in the HDR frame

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
torbsorb: I don't recall exactly either. However, I believe you argued that you want an example which shows how you can hard code all settings. I argued we needed an example which will read from .yml. Maybe the following can be a compromise:

CaptureHDRCompleteSettings.cpp will set settings in code, for one HDR (multiple frames).
CaptureHDRLoop.cpp will read a set of settings from .yml for each HDR in the loop. I can update the sample as I believe it should be.

I believe this could be clearer, and cover both of our wishes.

Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp
SatjaSivcev: done

SatjaSivcev:
@torbsorb I am moving this comment from code change.

CaptureHDRCompleteSettings.cpp will set settings in code, for one HDR (multiple frames).

CaptureHDRLoop.cpp will read a set of settings from .yml for each HDR in the loop. This sample will demonstrate three different HDR frames, each consisting of three frames. The three HDR images will have the same dynamic range (same coverage), but one will be the fastest, one will be the slowest but with the greatest quality.

How does this sound?

torbsorb:
@knatten, @SatjaSivcev, I rebased quite significantly. Now the history reads a natural evolution. The last commit doesn't strictly belong to this PR, but since it's a separate commit I guess it's ok?

SatjaSivcev:
I think it is ok. I made some comments. Will look again later.

README.md
```diff
@@ -17,21 +17,21 @@ There are two main categories of samples: **Camera** and **Applications**. The s
...
-    - **Visualization**
-      - [**CaptureHDRLoop**][CaptureHDRLoop-url] - This example shows how to acquire HDR images from the Zivid camera in a loop (while actively changing some HDR settings).
       - [**CaptureFromFileVis3D**][CaptureFromFileVis3D-url] - This example shows how capture a Zivid point cloud from file, and visualize it.
       - [**CaptureVis3D**][CaptureVis3D-url] - This example shows how to capture a Zivid point cloud, and visualize it.
       - [**CaptureLiveVis3D**][CaptureLiveVis3D-url] - This example shows how to continuosly capture a Zivid point cloud, and visualize it.
       - [**CaptureWritePCLVis3D**][CaptureWritePCLVis3D-url] - This example shows how capture a Zivid point cloud, save it to a .PCD file format, and visualize it.
       - [**ReadPCLVis3D**][ReadPCLVis3D-url] - This example shows how to read a PCL point cloud and visualize it.
     - **FileFormats**
-      - [**ReadIterateZDF**][ReadIterateZDF-url] - This example shows how to import and display a Zivid point cloud from a .ZDF file.
+      - [**ReadIterateZDF**][ReadIterateZDF-url] - This example shows how to import and iterate through a Zivid point cloud from a .ZDF file.
```
SatjaSivcev: For consistency with the description we accepted in https://github.com/zivid/matlab-samples/commit/ccb86a030de3c8c00a9ca2c696a9f1c505454ce5 modify to:

This example shows how to import a Zivid point cloud from a .ZDF file, iterate through, and extract individual points.

source/Camera/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,49 @@
...
+        {
+            std::cout << "Configure HDR settings" << std::endl;
+
+            std::vector<Zivid::Settings> settingsVector;
+            for(size_t frame = 1; frame <= number_of_frames_per_capture; frame++)
+            {
+                std::stringstream file_path;
+                file_path << "camera_settings/set" << set << "/frame_0" << frame << ".yml";
+                std::cout << "Add settings from " << file_path.str() << ":" << std::endl;
+                auto setting = Zivid::Settings::Settings(file_path.str());
```
SatjaSivcev: settings instead of setting

source/Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp
```diff
@@ -0,0 +1,60 @@
...
+            .set(Zivid::Settings::Filters::Gaussian::Sigma{ 1.5 })
+            .set(Zivid::Settings::Filters::Outlier::Enabled::yes)
+            .set(Zivid::Settings::Filters::Outlier::Threshold{ 5 })
+            .set(Zivid::Settings::Filters::Reflection::Enabled::yes)
+            .set(Zivid::Settings::Filters::Saturated::Enabled::yes)
+            .set(Zivid::Settings::BlueBalance{ 1.081 })
+            .set(Zivid::Settings::RedBalance{ 1.709 });
+
+        std::cout << "Configuring settings different for all HDR frames" << std::endl;
+        size_t iris[3] = { 10U, 20U, 30U };
```
SatjaSivcev: The frame exposure settings should be chosen wisely based on some criteria.

One idea is to have three different frames, all providing the same meaningful dynamic range, so the same coverage. One is to be the fastest, one the slowest but with the best quality.

There should be comments / print outputs that explain this. CA gives just one set of settings, this one shows an option to tune the output of the CA.

source/camera_settings/set1/frame_01.yml
```diff
@@ -0,0 +1,23 @@
+__version__: 3
+Settings:
+  Bidirectional: no
```
SatjaSivcev: Should these files be located at source/camera_settings/? If they are used just by this one sample, why not place them at  source/Camera/Basic/CaptureHDRLoop/CameraSettings/ ?

source/Camera/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,49 @@
...
+int main()
+{
+    try
+    {
+        Zivid::Application zivid;
+
+        std::cout << "Connecting to the camera" << std::endl;
+        auto camera = zivid.connectCamera();
+
+        const size_t number_of_captures = 3;
```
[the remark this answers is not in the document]
SatjaSivcev: Done.

README.md
SatjaSivcev: Done. Commit to be fixed up with rebase.

source/Camera/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
SatjaSivcev: Done. Commit to be fixed up with rebase.

SatjaSivcev:
@torbsorb I've pushed some minor changes. I've left my changes in separate commits that are to be fixed up if we agree to accept them. For other comments, we need to make decisions.

source/camera_settings/set1/frame_01.yml
torbsorb: I though maybe they'd be used by other samples, and also I wanted to separate source code from resources. In fact I would like to discuss maybe having a resources/ folder on the same level as source/. We could then put Zivid3D.zdf there too, and other files until we have a cloud storage...

SatjaSivcev: The resources idea seems good. Let's leave the directory structure as it is now.

However, please decide on the directory name so we can make it consistent between different repos.

torbsorb: I'm changing the folder name to /settings. I believe it's ok, since the camera settings are placed in subfolders (setX). Thus, it's ok that the name doesn't force <camera>-only settings in it.

torbsorb:
the last force-push was to pass CI

torbsorb:
last force push updates settings according to https://zividlabs.atlassian.net/browse/CS-266

source/Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp
SatjaSivcev: Implement one of these here: https://zividlabs.atlassian.net/browse/CS-266

This means, change values for iris and exposureTime arrays, and add a gain array as well.

SatjaSivcev:
It is good for me now. I am not a reviewer so I can't approve. @knatten can you do the last check so we can merge?

source/Camera/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,49 @@
...
+        {
+            std::cout << "Configure HDR settings" << std::endl;
+
+            std::vector<Zivid::Settings> settingsVector;
+            for(size_t frame = 1; frame <= framesPerCapture; frame++)
+            {
+                std::stringstream settingsPath;
+                settingsPath << "Settings/set" << set << "/frame_0" << frame << ".yml";
+                std::cout << "Add settings from " << settingsPath.str() << ":" << std::endl;
+                auto setting = Zivid::Settings(settingsPath.str());
```
knatten: Nitpick: Could be const

source/Camera/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,49 @@
+/*
+This example shows how to acquire HDR images from the Zivid camera in a loop
```
knatten: This should probably match the docs in the README?

source/CMakeLists.txt
```diff
@@ -144,3 +145,14 @@ endif()
...
 endif()
+
+add_custom_target(
+    CopyCameraSettings
+    COMMAND ${CMAKE_COMMAND} -E copy_directory
+            ${CMAKE_CURRENT_SOURCE_DIR}/Settings/
+            ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/Settings
+)
+if(TARGET CaptureHDRLoop)
+    message(STATUS "Copy resources for : ${TARGET}")
```
knatten: Leftover debug print? Don't think we need this, we're not using it for the other things we copy.

source/Camera/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,49 @@
...
+        Zivid::Application zivid;
+
+        std::cout << "Connecting to the camera" << std::endl;
+        auto camera = zivid.connectCamera();
+
+        const size_t captures = 3;
+        const size_t framesPerCapture = 3;
+        for(size_t set = 1; set <= captures; set++)
+        {
+            std::cout << "Configure HDR settings" << std::endl;
```
knatten: Use same mood all over? Previously you say "connecting", here it should probably be "Configuring".

source/Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp
```diff
@@ -0,0 +1,61 @@
...
+            .set(Zivid::Settings::Filters::Gaussian::Sigma{ 1.5 })
+            .set(Zivid::Settings::Filters::Outlier::Enabled::yes)
+            .set(Zivid::Settings::Filters::Outlier::Threshold{ 5 })
+            .set(Zivid::Settings::Filters::Reflection::Enabled::yes)
+            .set(Zivid::Settings::Filters::Saturated::Enabled::yes)
+            .set(Zivid::Settings::BlueBalance{ 1.081 })
+            .set(Zivid::Settings::RedBalance{ 1.709 });
+
+        std::cout << "Configuring settings different for all HDR frames" << std::endl;
+        size_t iris[3] = { 17U, 27U, 27U };
```
knatten: Nitpick: These arrays can be `const`.

Applications/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
knatten: @SatjaSivcev is correct, whenever you already have the full object, they are equivalent. `emplace_back` is faster if you only have the parts with which to construct the object.

source/Camera/Basic/CaptureHDRLoop/CaptureHDRLoop.cpp
```diff
@@ -0,0 +1,49 @@
...
+        for(size_t set = 1; set <= captures; set++)
+        {
+            std::cout << "Configuring HDR settings" << std::endl;
+
+            std::vector<Zivid::Settings> settingsVector;
+            for(size_t frame = 1; frame <= framesPerCapture; frame++)
+            {
+                std::stringstream settingsPath;
+                settingsPath << "Settings/set" << set << "/frame_0" << frame << ".yml";
+                std::cout << "Add settings from " << settingsPath.str() << ":" << std::endl;
```
knatten: This should probably be "Adding"?

knatten:
> The last commit doesn't strictly belong to this PR, but since it's a separate commit I guess it's ok?

Yeah I would have preferred a separate PR. In case we want to revert either that or the other things, it's better to be able to revert a full merge. But that's unlikely, so I don't mind for now.

PR #66, opened 2020-02-18
CI: Use correct source and build dirs

### LINKED ISSUE #65: Can't run lint.sh after build.sh
The linter finds generated files in the build folder. It should only look in the source folder.

Now that we've fixed #48, this should be an easy fix.

knatten:
`build.sh` used to build in the directory `/source/build` relative to the
repo root. Now we instead build in `/build` relative to the repo root.
This was the intention all along, the previous was just a bug.

`lint.sh` used to look for files in the entire repo. It now only looks in
the `source` folder, in order to avoid finding generated files in the
`build` folder.

This fixes #65


knatten:
Rebased on master

PR #69, opened 2020-02-25
Update license

SatjaSivcev:
I've taken the string from Zivid.h.
I've only changed support@zivid.com to customersuccess@zivid.com.
Should we remove the first sentence related to Zivid API?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2019, Zivid AS
+Copyright (c) 2020, Zivid AS
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without


SatjaSivcev:
@nedrebo can you check this?

SatjaSivcev:
@nedrebo reminder for license review

nedrebo:
Why do you think new license is required?

Have you seen this: https://github.com/zivid/zivid-licenses

That is the only place to fetch licenses from.

SatjaSivcev:
> 
> 
> Why do you think new license is required?
> 
> Have you seen this: https://github.com/zivid/zivid-licenses
> 
> Licensing is also covered here: https://github.com/zivid/zivid-repository-guidelines
> 
> That is the only place to fetch licenses from. It is important to copy the licenses verbatim, do not modify or add anything to the LICENSE file.

It was required because it stated 2019. I know what to do know.

PR #84, opened 2020-03-30
Add clang-tidy to continuous integration

torbsorb:


continuous-integration/linux/lint.sh
```diff
@@ -28,4 +28,17 @@ if [ $errorsFound -ne 0 ]; then
...
     exit 1
 fi
+
+BUILD_DIR="$ROOT_DIR/build/ci/tidy"
+mkdir --parents "$BUILD_DIR" || exit $?
+cd "$BUILD_DIR" || exit $?
+cmake -GNinja \
+    -DUSE_EIGEN3=OFF \
+    -DUSE_OPENCV=OFF \
+    -DCMAKE_CXX_CLANG_TIDY="/usr/bin/clang-tidy-8" \
```
nedrebo: I would bumpt to 9 and probably to 10 as well as it is going to be release before this PR is done.

.clang-tidy
```diff
@@ -0,0 +1,3 @@
+Checks: >
+  *,
```
nedrebo: A good starting point is to disable all warnings and then enable one warning again in follow-up commits, even if that warning has zero or just one hit. That makes the reviewing process easy. Also it makes it possible to merge to master before everything is fixed.

To start this I would run clang-tidy using current config, then whenever you get an error you add a `-test-name` on a new line like you you see here: https://github.com/zivid/zivid-python/blob/master/.clang-tidy

Then, once you pass, you have a baseline for you first commit, with many warnings disabled. Then start to peel of suppressions again until you at least are as good as we are in `zivid-python`. If you hit some warnings that is hard to fix, weird or that feel counterproductive I suggest you post in the Programming channel on Teams and @me. I will monitor it and the discussions may be of interest to others.

nedrebo: You can just immediately disable:
* -fuchsia-*: These tests are for embedded code
* -llvm-header-guard: We use `#pragma once`
* -google-runtime-references: We disagree with that one. References are not confusing.

PR #85, opened 2020-03-31
Add clang-tidy to continuous integration

torbsorb:


torbsorb:
clang-tidy throws a few warnings that I'm not sure if we should fix, or ignore

nedrebo:
Can you elaborate on the which warnings you are talking about and links to the code they triggers on?

torbsorb:
> 
> 
> Can you elaborate on the which warnings you are talking about and links to the code they triggers on?

In particular use of "magic" numbers in Zivid Settings initialization, such as:
https://github.com/zivid/cpp-extra-samples/blob/fb211e2b8daa8fcac2a0a4b995214952fe1865c8/source/Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp#L26

throws https://dev.azure.com/zivid-devops/cpp-extra-samples/_build/results?buildId=3049&view=logs&j=78522f12-3432-53d9-a26c-09281b784bba&t=00999172-2923-5109-b879-17c38f1330bd&l=5789

Also `error: do not use array subscript when the index is not an integer constant expression; use gsl::at()` as in (line 45)
https://github.com/zivid/cpp-extra-samples/blob/fb211e2b8daa8fcac2a0a4b995214952fe1865c8/source/Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp#L41-L45

nedrebo:
I would disable the magic number test and just not use C-arrays at all.

nedrebo:
You can use vector or array and the `.at()` member function that do bounds checking.

torbsorb:
This is now clean in local CI.

Some decisions have been discussed on Teams, for example disabling magic-number rule.
Two rules are disabled that I didn't want to:
`-cppcoreguidelines-pro-bounds-pointer-arithmetic,`
* disabled because CameraUserData (basic sample copied from SDK) uses `argv`. An alternative is to use Clipp. I have another PR for that. It seems a bit excessive to ask customers to download the .hpp file and add path. I wonder if we can simply host it.

`-cppcoreguidelines-pro-type-union-access,`
* PCL library has defined it's point access a bit unfortunate. This causes to use union member to access the variable. We could probably rewrite it and compose the float it maps to, but that code is not going to be pretty.

SatjaSivcev:
> This is now clean in local CI.
> 
> Some decisions have been discussed on Teams, for example disabling magic-number rule.
> Two rules are disabled that I didn't want to:
> `-cppcoreguidelines-pro-bounds-pointer-arithmetic,`
> * disabled because CameraUserData (basic sample copied from SDK) uses `argv`. An alternative is to use Clipp. I have another PR for that. It seems a bit excessive to ask customers to download the .hpp file and add path. I wonder if we can simply host it.
> 
> `-cppcoreguidelines-pro-type-union-access,`
> * PCL library has defined it's point access a bit unfortunate. This causes to use union member to access the variable. We could probably rewrite it and compose the float it maps to, but that code is not going to be pretty.

Is it possible to ignore these per code line for the lines that cause them (and then enable in general)? Would that be ugly?

torbsorb:
Just pushed a commit that allows us to bring back `cppcoreguidelines-pro-bounds-pointer-arithmetic`. We can probably do `//NOLINT` for the union usage errors. Was only a handful number of lines

torbsorb:
Suggestion:
```cpp
for(size_t i = 0; i < cloud.points.size(); ++i)
{
    auto point = pointCloud.operator()(i);

    cloud.points[i].x = point.x;       // NOLINT(cppcoreguidelines-pro-type-union-access)
    cloud.points[i].y = point.y;       // NOLINT(cppcoreguidelines-pro-type-union-access)
    cloud.points[i].z = point.z;       // NOLINT(cppcoreguidelines-pro-type-union-access)
    cloud.points[i].r = point.red();   // NOLINT(cppcoreguidelines-pro-type-union-access)
    cloud.points[i].g = point.green(); // NOLINT(cppcoreguidelines-pro-type-union-access)
    cloud.points[i].b = point.blue();  // NOLINT(cppcoreguidelines-pro-type-union-access)
}
```

source/3rd-party/clipp/include/clipp.hpp
```diff
@@ -0,0 +1,6408 @@
+/*****************************************************************************
+ *  ___  _    _   ___ ___
+ * |  _|| |  | | | _ \ _ \   CLIPP - command line interfaces for modern C++
+ * | |_ | |_ | | |  _/  _/   version 1.1.0
```
nedrebo, line 4: Latest version is 1.2.3: https://github.com/muellan/clipp/blob/master/include/clipp.h

torbsorb: done

torbsorb:
Branch CI is now clean. If we complete the review then I can rebase and fix the final problems.

nedrebo:
I am not sure if you are waiting for me or plan to do more before this is ready for review again. If you are waiting for me let me know.
This branch is out of date, update it before next review round.

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -16,15 +16,17 @@ namespace
 
     Duration computeMedianDuration(std::vector<Duration> durations)
     {
+        Duration result;
         std::sort(durations.begin(), durations.end());
         if(durations.size() % 2 == 0)
         {
-            return (durations.at(durations.size() / 2 - 1) + durations.at(durations.size() / 2)) / 2;
+            result = (durations.at(durations.size() / 2 - 1) + durations.at(durations.size() / 2)) / 2;
```
nedrebo, line 23: No, just remove the else, you should never have wrongly initialized data, ref `Duration result`.

Strictly speaking the program state is wrong between line 19 and 23, that is not good if people not paying attention to that refactor this function, suddenly this function returns the uninitialized value. This is btw a reason to use `const` and/or `auto`, because then you cannot get into that mess. But in this situation just return and you have one variable less and less program state that can be wrong.

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -226,10 +228,13 @@ namespace
 
     size_t getMinExposureTime(const std::string &modelName)
     {
+        size_t exposureTime;
```
nedrebo, line 231: Same here. Looking at those two codes (before and after) it is clearly better before.

You should not change to something worse just because a tool tells you to.

In that case you need to shout and scream, ether you find a way to improve the code and if no one can tell you a better way or explain why we want this warning, then the diagnostics should be disabled.

torbsorb: I didn't like it, but thought maybe you wanted the else..
I almost proposed the following:
```cpp
        size_t exposureTime = exposureTime = 8333; // Min for Zivid One
        if(modelName.substr(0, 14) == "Zivid One Plus")
            exposureTime = 6500; // Min for Zivid One Plus

        return exposureTime;
```

but, I'll take your change.

continuous-integration/linux/lint.sh
```diff
@@ -28,4 +28,17 @@ if [ $errorsFound -ne 0 ]; then
     echo "ERROR: $errorsFound formatting error(s) found! See the diffs for each file printed above."
     exit 1
 fi
+
+BUILD_DIR="$ROOT_DIR/build/ci/tidy"
+mkdir --parents "$BUILD_DIR" || exit $?
+cd "$BUILD_DIR" || exit $?
+cmake -GNinja \
+    -DUSE_EIGEN3=OFF \
```
nedrebo, line 36: Why off? Don't we lose out on testing of many files?

.clang-tidy
```diff
@@ -0,0 +1,401 @@
+Checks: >
+  *,
```
nedrebo, line 2: I would have fixed:
  -bugprone-suspicious-semicolon,
  -hicpp-braces-around-statements,
  -google-readability-braces-around-statements,
  -readability-misleading-indentation,
  -bugprone-macro-parentheses,

PR #86, opened 2020-04-10
Add assisted capture to ZividBenchmark

SatjaSivcev:
This adds the training phase of capture assistant (suggest settings).

### DIFF
--- a/source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
+++ b/source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
@@ -132,6 +132,11 @@ namespace
         if(!filterList.empty()) std::cout << "  filters = " << filterList << std::endl;
     }
 
+    void printAssistedCapture3DHeader(const size_t numFrames)
+    {
+        printHeaderLine(numFrames, { "Running assisted capture ", " times:" });
+    }
+
     void printCapture2DHeader(const size_t numFrames, const Zivid::Settings2D &settings)
     {
         printHeaderLine(numFrames, { "Capturing ", " 2D frames:" });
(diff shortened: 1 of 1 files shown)


SatjaSivcev:
Please review @torbsorb @knzivid

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -317,6 +327,36 @@ namespace
...
+    {
+        printAssistedCapture3DHeader(numFrames);
+
+        Zivid::CaptureAssistant::SuggestSettingsParameters suggestSettingsParameters(
+            std::chrono::milliseconds{ 1200 }, Zivid::CaptureAssistant::AmbientLightFrequency::none);
+
+        for(size_t i = 0; i < 5; i++) // Warmup
+        {
+            const auto settingsVector{ Zivid::CaptureAssistant::suggestSettings(camera, suggestSettingsParameters) };
+        }
```
torbsorb, line 340: is it required to run warmup on this specifically? If so, why is it sufficient?

knzivid:
I do not have any comments on this PR, the changes look good to me

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
SatjaSivcev: The first capture is always much slower here, see the image below. The second capture can be slightly slower (e.g. 880 for 820 average), but this one I saw only once in 15-ish tests.

The observation is similar for 3D captures, but for them, occasionally, a number of first 5 frames can be outliers also (this happens more commonly than above).

I wanted to keep this warmup consistent with the ones in other functions where we use x5, benchmarkCapture2D() is the most similar case to this one and it also uses 5.

Alternatively, we could use 1 or 2 warmup captures for 2D captures and for this. What do you think?

image

PR #89, opened 2020-04-21
Benchmark raw output

torbsorb:
We want to be able to collect all the data from each measurement, not just the mean and median. Design criteria:
- No external libraries (tempting to use yaml-cpp)
- Very minor modification to existing code (tempting to refactor completely)

Added two unrelated commits before the most important commit

### DIFF
--- a/source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
+++ b/source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
@@ -1,10 +1,12 @@
 #include <Zivid/Zivid.h>
 #include <algorithm>
+#include <fstream>
 #include <iostream>
 
 namespace
 {
     const int PRINT_WIDTH = 56;
+    const char *RAW_FILE_NAME = "zivid_benchmark_raw.txt";
 
     using HighResClock = std::chrono::high_resolution_clock;
     using Duration = std::chrono::nanoseconds;
@@ -72,6 +74,22 @@ namespace
         return {};
     }
 
+    void createEmptyOutputFile()
+    {
+        std::ofstream rawFile(RAW_FILE_NAME, std::ios::out);
+    }
+
+    void addToRawFile(const std::string &string)
+    {
+        std::ofstream rawFile(RAW_FILE_NAME, std::ios::out | std::ios::app);
+        rawFile << string << std::endl;
+    }
+
+    void addStringsToRawFile(const std::vector<std::string> &strings)
+    {
+        std::for_each(strings.cbegin(), strings.cend(), addToRawFile);
+    }
+
     void printProgress(const int iteration, const int total)
     {
         std::cout << "..." << iteration << "/" << total << "   "
@@ -126,26 +144,36 @@ namespace
     void printConnectHeader(const size_t numConnects)
     {
         printHeaderLine(numConnects, { "Connecting and disconnecting ", " times each (be patient):" });
+        addToRawFile("Connecting and disconnecting ");
     }
 
     void printCapture3DHeader(const size_t numFrames, const std::vector<Zivid::Settings> &settingsVector)
     {
         const auto filterList = makefilterList(settingsVector);
         printHeaderLine(numFrames, { "Capturing ", " 3D frames:" });
+        addToRawFile("Capturing 3D frames:");
         std::cout << "  exposure time = " << makeSettingList<Zivid::Settings::ExposureTime>(settingsVector)
                   << std::endl;
+        addToRawFile("  exposure time = " + makeSettingList<Zivid::Settings::ExposureTime>(settingsVector));
         std::cout << "  iris settings = " << makeSettingList<Zivid::Settings::Iris>(settingsVector) << std::endl;
-        if(!filterList.empty()) std::cout << "  filters = " << filterList << std::endl;
+        addToRawFile("  iris settings = " + makeSettingList<Zivid::Settings::Iris>(settingsVector));
+        if(!filterList.empty())
+        {
+            std::cout << "  filters = " << filterList << std::endl;
+            addToRawFile("  filters = " + filterList);
+        }
     }
 
     void printAssistedCapture3DHeader(const size_t numFrames)
     {
         printHeaderLine(numFrames, { "Running assisted capture ", " times:" });
+        addToRawFile("Running assisted capture:");
     }
 
     void printCapture2DHeader(const size_t numFrames, const Zivid::Settings2D &settings)
     {
         printHeaderLine(numFrames, { "Capturing ", " 2D frames:" });
+        addToRawFile("Capturing 2D frames:");
         std::cout << "  exposure time = { " << settings.exposureTime().toString() << " }" << std::endl;
     }
 
@@ -215,9 +243,13 @@ namespace
     void printZividInfo(Zivid::Camera camera)
     {
         std::cout << "API: " << Zivid::Version::libraryVersion() << std::endl;
+        addToRawFile("API: " + Zivid::Version::libraryVersion());
         std::cout << "OS: " << OS_NAME << std::endl;
+        addToRawFile("OS: " + std::string(OS_NAME));
         std::cout << "Camera: " << camera << std::endl;
+        addToRawFile("Camera: " + camera.toString());
         std::cout << "Compute device: " << camera.computeDevice() << std::endl;
+        addToRawFile("Compute device: " + camera.computeDevice().toString());
         printPrimarySeparationLine();
         printCentered("Starting Zivid Benchmark");
     }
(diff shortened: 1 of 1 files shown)


source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -120,26 +139,36 @@ namespace
...
     }
 
     void printCapture3DHeader(const size_t numFrames, const std::vector<Zivid::Settings> &settingsVector)
     {
         const auto filterList = makefilterList(settingsVector);
         printHeaderLine(numFrames, { "Capturing ", " 3D frames:" });
+        addToRawFile("Capturing 3D frames:");
         std::cout << "  exposure time = " << makeSettingList<Zivid::Settings::ExposureTime>(settingsVector)
                   << std::endl;
+        addToRawFile("  exposure time = " + makeSettingList<Zivid::Settings::ExposureTime>(settingsVector));
```
SatjaSivcev, line 152: You add exposure time here but you don't add it at line 171 in capture2D. Choose one, my vode goes to minimalism.

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -304,14 +341,17 @@ namespace
...
             const auto beforeCapture = HighResClock::now();
             const auto frame = Zivid::HDR::capture(camera, settingsVector);
             const auto afterCapture = HighResClock::now();
             frame.getPointCloud();
             const auto afterProcess = HighResClock::now();
 
             captureDurations.push_back(afterCapture - beforeCapture);
+            addToRawFile(std::string("capture duration: ") + formatDuration(afterCapture - beforeCapture));
             processDurations.push_back(afterProcess - afterCapture);
+            addToRawFile(std::string("process duration: ") + formatDuration(afterProcess - afterCapture));
```
SatjaSivcev, line 354: Can we add '3D' here to make it consistent with 2D capture?

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
torbsorb: I'll add it to both, we need to for 3D to distinguish the data

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
torbsorb: sure :)

torbsorb:
> 
> 
> There are just two minor comments from me.

I addressed those, right?

SatjaSivcev:
> 
> 
> > There are just two minor comments from me.
> 
> I addressed those, right?

Yes

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -72,6 +74,23 @@ namespace
         return {};
     }
 
+    void createEmptyOutputFile()
```
nedrebo, line 77: This should return the file handle, opening an closing the file many times is inefficient and not very elegant.

torbsorb: even though it means passing the file handle into quite nice functions, that doesn't seem to need them? In addition to the verbosity enum? I could make those changes and push as fixup. 

However, I would rather refactor the entire program, and have those nice benchmark functions return the raw results. We can for example add one layer like this (one example):
Existing:
- `benchmarkConnect(camera, numConnects);`

Proposal:
- 'benchmarkConnect(camera, numConnects, fileHandle, verbosityEnum);'
  - `printHeaderBenchmarkConnect(fileHandle, verbosityEnum)`
  - `results = benchmarkConnect(camera, numConnects)`
  - `processBenchmarkConnectResults(results fileHandle, verbosityEnum)`

I would then pull out any processing of the results from `benchmarkConnect(camera, numConnects)`

nedrebo: If you do the verbose cli options my proposal is to not deal with files in here at all, just dump all to std out. Then this function goes away.

To answer your question, if you were to keep this function, I would have created a Printer class that opens the file in ctor and closes it in dtor and then all the nice functions you talk about would be member functions on that class.

nedrebo: To avoid passing the verbosity level around you could still make the Printer class, but not pass it a file name, but instead pass it the verbosity level in ctor. Then all functions can be moved into that class and signatures can be kept unchanged.

A printer class also makes it easier to add more customizations points later on.

torbsorb: > 
> 
> If you do the verbose cli options, then my proposal is to not deal with files in this program at all, just dump everything to stdout. This function then goes away.

The idea was to allow customer to see the nice output, while we could still ask for the raw data. With your option he would have to run it twice, or be bothered with the verbose prints.

torbsorb: @SatjaSivcev, what do you think about just writing to std::cout. I'm thinking we would only print what's currently going to the file when the verbosity is enabled. In other words it would be an OR operation. The instruction for the customer could then be to run 
```bash
./ZividBenchmark.exe --vv > zivid_benchmark_raw.txt
```

nedrebo: One minus btw. `-vv` is shorthand for `-v -v` which is shorthand for `--verbose --verbose`. Just convention I have seen some places. Feel free to call it other things.

SatjaSivcev: The program would output raw data you add `-vv`? And if you don't, it would output just the mean and median as of now?

SatjaSivcev: Running twice or verbose prints is something I would avoid. If possible, I would avoid options, it is best if the customer has only one way to run the program and that gives them a nice print and us raw data. Can this be achieved with the above suggestion?

torbsorb: I believe a compromise would be that we tag the data and use `| grep` or `| Select-string` (for Powershell). What do you think about that? This could be combined with a verbose option, so that the default execution only gives a nice print.

nedrebo: Only with option @SatjaSivcev, but then yes. Tools dumping debug files to $PWD without me asking for it would annoy me a lot at least.

Keep it simple, no grepping, no magic. Verbose mode (`--verbose`) is very common or option to save file (`--save-debug-info file-name`) would be the two options that I feel is natural.

Think I sated my opinion. Others view on this? @knatten, @apartridge, @knzivid

knzivid: I agree with not touching the home directory. Running the benchmark twice would have counter-intuitive behavior (be it overwriting or appending to the existing file)

torbsorb: @knzivid, thanks for your feedback! You seem to argue against running the benchmark twice, like we do. Yet I'm not sure if you argue against a `--verbose` option. You also seem to argue against `--save_to_file`. May I ask you to refine your opinion?

A `--verbose` option doesn't give a nice output, hence why we say it would lead to multiple runs.

A `--save-to-file` option seems like the best alternative to me, as it 
- allows for nice output in console
- gives us the raw data
- makes the customer aware that we're writing to $PWD

I did not receive any feedback on the data-tagging option. It's quite common to have [INFO], [WARNING] and [ERROR] in logs, right? I was thinking about only tagging the data-output, thus making it possible to run `--verbose` without actually seeing it. I mentioned grep, but it seems awk can do the job better in splitting the output. I'm not a fan of this option btw, I'd prefer `--save-to-file`

nedrebo: My response to "data-tagging" was "Keep it simple, no grepping, no magic.".

torbsorb: Ok, so we'll go for the `--save-to-file` option.

nedrebo: Would call it, `--save-debug-info` or similar since it does not save the regular output.

PR #97, opened 2020-04-24
Add file camera to the repository

### LINKED ISSUE #94: No instructions for running CaptureFromFile on Ubuntu about the file camera
Ubuntu installation doesn't provide the file camera. There are no instructions that it is required to run this code sample and where to download it from. Only the following error message appears: Error: Test data directory not found. Please set environment variable ZIVID_DATA.

SatjaSivcev:
This commit simplifies running file camera samples. For Ubuntu, it was required to download the file camera (MiscObjects.zdf) from the Zivid website.

Fixes #94

### DIFF
--- a/source/Applications/Basic/Visualization/CaptureFromFileVis3D/CaptureFromFileVis3D.cpp
+++ b/source/Applications/Basic/Visualization/CaptureFromFileVis3D/CaptureFromFileVis3D.cpp
@@ -16,7 +16,7 @@ int main()
         Zivid::CloudVisualizer vis;
         zivid.setDefaultComputeDevice(vis.computeDevice());
 
-        auto zdfFile = Zivid::Environment::dataPath() + "/MiscObjects.zdf";
+        auto zdfFile = "MiscObjects.zdf";
 
         std::cout << "Initializing camera emulation using file: " << zdfFile << std::endl;
         auto camera = zivid.createFileCamera(zdfFile);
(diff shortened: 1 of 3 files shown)


nedrebo:
I am not sure it is wise to duplicate this. It sign CS up for recurring additional maintenance efforts and adds the risk of things getting out of sync.

SatjaSivcev:
> It sign CS up for recurring additional maintenance efforts and adds the risk of things getting out of sync.

What is the alternative? We already do this for the other repos, e.g. python-samples. I wish all sample repos could automatically download files from one source, and we want to investigate that approach. For now, I wanted to make all sample repos consistent and also user friendly because as of now there are no instructions on how to get MiscObject.zdf. We could add those instructions bu that is an unnecessary extra effort for the user.

nedrebo:
If you are not happy with how we distribute the file camera you should report it as a defect in the SDK so SE can fix it. We have a longer iteration time than you, but SDK *is* the right place to improve. If we disagree with it being a defect you must improve documentation so customers can deal with what we have. Duplicating files in this repository is just band a band aid for a specific case.

On a general note: Fixing issues at this level means you need to "fix" it in all you repositories and still the issue is not solved customers using only the SDK. This is a bad spiral, by hacking around things in the SDK and adding maintenance cost at this level, you need to hire more people to maintain it eventually. These resources should rather be added to the SE team so we can make a better product in the first place.

It is very important that CS is a demanding "customer" and require improvements in SDK and not accept that you need to hack around things. You should be our most demanding customer so that you pave the way for improvements in SDK before real customers are affected by it. If you hack around things without us knowing we cannot improve. If you have resource to do things like this, I rather have it that you provide patches to SDK. That will require more effort from you, but the improvement will be generally available, alway in sync and the maintenance cost will be on the SE team.

PR #101, opened 2020-04-30
Fix OpenCV version

### LINKED ISSUE #74: README/CMakeLists.txt and Knowledge Base don't agree on OpenCV version
Both `REAMDE.md` and `CMakeLists.txt` say we require OpenCV 4.0.1. The [knowledge base article](https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/61472793/Configure+C+Extra+Samples+with+CMake+and+build+them+using+Visual+Studio+in+Windows) linked to from `README.md` says 4.1.1.

I would assume 4.0.1 is the correct one, and the KB has to be updated, but I'm not sure.

SatjaSivcev:
Fixes #74 

The correct version is 4.1.0. actually (not 4.0.1. or 4.1.1)

KB is fixed: https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/61472793

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,10 @@ There are two main categories of samples: **Camera** and **Applications**. The s
         - [Eigen](http://eigen.tuxfamily.org/) version 3.3.7 or newer
     - [**CaptureUndistortRGB**][CaptureUndistortRGB-url] - Use Zivid camera intrinsics to undistort an RGB image. This example will prompt the user for whether to capture a 2D or a 3D image. In both instances it will operate on a 2D image. However, in the 3D case it will extract 2D image from a ZDF point cloud. The 2D variant is faster.
       - **Dependencies:**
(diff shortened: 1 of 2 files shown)


knatten:
The KB still says

> The OpenCV version that is tested is 4.1.0.

Can you update that before I approve this PR? Just to make sure we don't forget.

knatten:
Sorry, I got it the wrong way around. 4.1.0 is the correct version! :)

torbsorb:
Sorry for being late to the party on this one. Ubuntu 20.04 installs 4.2.0. Should we just update to that?

knatten:
I guess that depends on how easy it is for customers to get 4.2.0 compared to 4.1.0 on various systems, and how much of a hassle it will be for existing customers to upgrade. Then weigh that against which features in 4.2.0 will make our samples how much better.

PR #114, opened 2020-06-04
Dont skip clang

knatten:
As of Zivid 2.0 we don't need to skip clang on Ubuntu 16.04. Also remove an outdated TODO.


knatten:
Yeah if you time, thanks!

PR #130, opened 2020-07-03
Update settings in Capture.cpp

ajlandau:
Update comments for ease of use. 
Change of settings in Capture2D.
Require user input to close the cmd window when an error occurs.


source/Camera/Basic/CaptureAssistant/CaptureAssistant.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-This example shows how to use Capture Assistant to capture point clouds, with color, from the Zivid camera.
```
SatjaSivcev, line 2: Remember to change this in other sample repos when you start 
image

If you think this is important it should be modified in all samples that have it and also in the README.md. Then it should be also modified in the other repos. I personally think it would be ok to skip it. I think both versions are gramatically correct.

ajlandau: I think you're right. We can just leave the comma in.  It is just a personal style choice.

source/Camera/Basic/CaptureWithSettingsFromYML/CaptureWithSettingsFromYML.cpp
```diff
@@ -1,6 +1,7 @@
 /*
 This example shows how to capture point clouds, with color, from the Zivid camera,
-with settings from YML file.
+with settings from YML file. The YML files for this samples can be found at:
+https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/450363393/Sample+Data
```
SatjaSivcev, line 4: We should avoid hard coding these kinds of instructions in the code samples. This is already covered in README.md. The reason to avoid doing this: if we change the download link, we have to change it in every sample that has it. We want to have only one place where we change it ideally.

If you want to emphasize, do it by saying that this is covered in the instructions.

image

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -1,5 +1,9 @@
-#include <Zivid/Zivid.h>
+/*
+Zividbenchmarks is a sample that will test the average speed of different operations on your computer. It will provide
```
SatjaSivcev, line 2: - Zivid**B**enchmark - capital B
- _the average speed_ - can we change this to _performance_? Emphesize that it is performance of the computer on using Zivid camera, not performance in general.
- _the mean and median_ - add _time_ - I would maybe even skip saying it is median and mean, time is sufficient I think. Maybe we add min, max, st. dev at some point, etc. no need to go into details in description.
- Let's try to use the same terms that we use in printing, e.g. _breakdown of the total capture time for single and HDR captures with and without filters_ instead of _single imaging, HDR and filtering._
We also measure _saving_ point clouds.

source/Camera/Basic/CaptureWithSettingsFromYML/CaptureWithSettingsFromYML.cpp
ajlandau: Updated with link removal.

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
ajlandau: Changed the wording and phrasing to match closer what we are doing in the code.

source/Camera/Basic/CaptureWithSettingsFromYML/CaptureWithSettingsFromYML.cpp
```diff
@@ -1,7 +1,7 @@
 /*
 This example shows how to capture point clouds, with color, from the Zivid camera,
-with settings from YML file. The YML files for this samples can be found at:
-https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/450363393/Sample+Data
+with settings from YML file. The YML files for this samples can be found under the main instructions
```
SatjaSivcev, line 3: Can you apply this change to the other samples that use sample data?
Hint: Open the repo folder in Visual Studio Code and search for ZIVID_SAMPLE_DATA_DIR - all samples should be listed

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -1,6 +1,9 @@
 /*
-Zividbenchmarks is a sample that will test the average speed of different operations on your computer. It will provide
-the mean and median for connects, disconnects, single imaging, HDR and filtering.
+ZividBenchmark is a sample that will test the performance of different operations on your computer. It will provide
```
SatjaSivcev, line 2: There are extra spaces and we can align the text. Also, we should use simple present tense, you can use this:

ZividBenchmark is a sample that tests the performance of different SDK operations on your computer. It provides measured time
in milliseconds for: connecting and disconnecting the camera, capture assistant, single and HDR captures with and without
filters, as well as saving point clouds.

SatjaSivcev:
Please address the latest comments.

source/Camera/Basic/CaptureWithSettingsFromYML/CaptureWithSettingsFromYML.cpp
SatjaSivcev: @ajlandau have you seen this comment?

SatjaSivcev: Also, there is a typo -> The YML files for this sample**s** - should be singular

ajlandau: Added the last sentence to all of the files.  fixed plural to singular. For the files that use other data in the sample data folder, changed what they were finding.
yml -> zdf
yml -> yaml  
to make it valid for the specific sample, is that okay?

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -139,6 +154,12 @@ int main()
     {
         Zivid::Application zivid;
 
+        // camera settings for both 2D and 3D acquisitions set
+        const std::chrono::microseconds exposureTime = std::chrono::microseconds{ 20000 };
+        const double aperture = 5.66;
+        const double gain = 1.0;
+        const double brightness = 1.0;
```
SatjaSivcev, line 161: Declare and initialize variables just before using them, meaning, move them just before:

`const auto bgr = (use2D ? getImage2D(camera, makeSettings2D(exposureTime, aperture, gain, brightness))
                                : getImage3D(camera, makeSettings(exposureTime, aperture, gain, brightness)));

`

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -139,6 +154,12 @@ int main()
     {
         Zivid::Application zivid;
 
+        // camera settings for both 2D and 3D acquisitions set
```
SatjaSivcev, line 157: I don't think we need this comment.

SatjaSivcev:
Two more changes requested @ajlandau

ajlandau:
Done and ready for review

SatjaSivcev:
Few comments on commit messages:
- _Update settings to improve dark acquisitions for samples_ - this is not true for all the samples, for example in Capture.cpp we change exposure time from 8333 to 6500. This change does not help with dark scenes (quite the opposite) so the comment can can be misleading. Make the commit description more 'generic' (e.g. improve default settings for generic scenes) or skip it.
- _Adds wait for user input on error message for samples_ - Change **Adds** to **Add** (Example: Do thing in imperative mood)

ajlandau:
Updated commit messages to coincide with standard format.

source/Camera/Basic/Capture2D/Capture2D.cpp
```diff
@@ -20,10 +20,10 @@ int main()
...
         const auto settings2D =
             Zivid::Settings2D{ Zivid::Settings2D::Acquisitions{ Zivid::Settings2D::Acquisition{
-                                   Zivid::Settings2D::Acquisition::ExposureTime{ std::chrono::microseconds{ 10000 } },
-                                   Zivid::Settings2D::Acquisition::Aperture{ 2.83 },
-                                   Zivid::Settings2D::Acquisition::Brightness{ 1.0 },
-                                   Zivid::Settings2D::Acquisition::Gain{ 1.0 } } },
+                                   Zivid::Settings2D::Acquisition::ExposureTime{ std::chrono::microseconds{ 30000 } },
+                                   Zivid::Settings2D::Acquisition::Aperture{ 11.31 },
+                                   Zivid::Settings2D::Acquisition::Brightness{ 1.80 },
+                                   Zivid::Settings2D::Acquisition::Gain{ 2.0 } } },
```
runenordmo, line 26: The [Capture.cpp](https://github.com/zivid/zivid-cpp-samples/blob/64862fa74d67153a4e1eca4ca70bb64efbcd7cdd/source/Camera/Basic/Capture/Capture.cpp#L21) uses the default gain equal to 1.0. Why is it an improvement to use gain equal to 2.0 here?

source/Camera/Basic/Capture/Capture.cpp
```diff
@@ -20,7 +20,7 @@ int main()
         const auto settings =
             Zivid::Settings{ Zivid::Settings::Acquisitions{ Zivid::Settings::Acquisition{
                                  Zivid::Settings::Acquisition::Aperture{ 5.66 },
-                                 Zivid::Settings::Acquisition::ExposureTime{ std::chrono::microseconds{ 8333 } } } },
+                                 Zivid::Settings::Acquisition::ExposureTime{ std::chrono::microseconds{ 6500 } } } },
```
runenordmo, line 23: Could you mention in the commit message why each of the settings you use here are considered preferable?

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -150,7 +165,13 @@ int main()
             use2D = true;
         }
 
-        const auto bgr = (use2D ? getImage2D(camera) : getImage3D(camera));
+        const std::chrono::microseconds exposureTime = std::chrono::microseconds{ 20000 };
```
runenordmo, line 168: Consider using `const auto` for all these variables.

source/Camera/Basic/Capture2D/Capture2D.cpp
SatjaSivcev: We want low iris (high f-number) to have the 2D image in focus. For that reason, we need to crank up other settings to get high enough exposure. Gain does not affect 2D images the same as it affects 3D images so there is no reason to be restrictive about using it.

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -187,6 +187,10 @@ int main()
     catch(const std::exception &e)
     {
         std::cerr << "Error: " << Zivid::toString(e) << std::endl;
-        return EXIT_FAILURE;
+        std::cout << "Press enter to exit." << std::endl;
+        if(std::cin.get() == '\n')
+        {
+            return EXIT_FAILURE;
+        }
```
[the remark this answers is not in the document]
SatjaSivcev: If you double click on the exe file on Windows the terminal will close and the user will not see the message. This will be perceived as though the program doesn't work.
This will not happen if you start the program from terminal, but most people don't start programs by opening the terminal but by double clicking on the exe file.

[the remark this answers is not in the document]
ajlandau: With how it is stated in the tutorials the samples should be able to run with just access the release folder and double-clicking on the generated .exe. With that mindset, I thought adding in the pause for just printing errors might be smart.

source/Camera/Basic/Capture/Capture.cpp
ajlandau: Changed to const auto

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
```diff
@@ -150,7 +165,13 @@ int main()
             use2D = true;
         }
 
-        const auto bgr = (use2D ? getImage2D(camera) : getImage3D(camera));
+        const std::chrono::microseconds exposureTime = std::chrono::microseconds{ 20000 };
+        const double aperture = 5.66;
+        const double gain = 1.0;
+        const double brightness = 1.0;
+
+        const auto bgr = (use2D ? getImage2D(camera, makeSettings2D(exposureTime, aperture, gain, brightness))
```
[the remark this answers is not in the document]
ajlandau: Taken care of :)

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
ajlandau: changed variables in main to const auto

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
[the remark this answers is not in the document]
runenordmo: This gives quite some duplication in all samples? What convention do you have for avoiding duplication within these samples?

It could be put in a function that is used by all samples. Don't know if this might interfere with each sample being standalone? @SatjaSivcev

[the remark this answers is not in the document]
SatjaSivcev: We don't want to have complicated code. Each sample should be standalone and simple. Samples are not SDK so there is no need to use one function (file) for something like this by all samples. I don't think this as of now is too much code.
The intention of the samples is for the customers to learn how to use the SDK functions. We don't want them to spend time figuring out what the code does.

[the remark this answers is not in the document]
runenordmo: > We don't want to have complicated code. Each sample should be standalone and simple. Samples are not SDK so there is no need to use one function (file) for something like this by all samples. I don't think this as of now is too much code.

I don't agree that it is more complicated. But definetly agree that it's a benefit that the sample is standalone (copy-paste-able); so if we want this user input on error I think we can have it like this 👍

runenordmo:
Looks good to me. If you rebase and update the commit message on why the new chosen settings are preferable 👍

source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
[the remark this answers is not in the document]
nedrebo: I agree this is weird/unstandard behavior and only complicated the source code, any CLI tool behaves this way if you double click it in the file explorer, that is just user error. I think it is better to update the documentation.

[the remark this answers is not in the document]
nedrebo: I assume KB uses VS as an example for how to build on Windows? Then just say press "Ctrl+F5" to run the sample. (Ref: https://stackoverflow.com/a/1775870)

If KB documents show how to build on the console on Windows then just say "./build/path/sample.exe"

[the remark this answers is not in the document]
SatjaSivcev: You assume that every user (including those who double click by default) will read the documentation. I don't think they will. So for those who won't, the program will seem not to work and they will not see the error message (because they will double click the exe file).

Can you explain what is the benefit of the previous solution, for developers? I don't get it, except that it has 4 lines of code less? I can understand that it is standard use but imp it has limitations.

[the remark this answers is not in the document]
nedrebo: IRL update: I think we mostly disagree because we have a different mental model of the end-user.

To make decisions about whether to do things like this in code and whether to spell out every tedious detail in KB I think we need two things.

1. One to three reference end-users with a particular skill set documented in Confluence that we optimize documentation experience for. These users can be used to align all documentation writers to the same mental model. and be referred to when having a dispute. Maybe also a couple of customer we do not optimize for (the expert and the one that cannot program at all, catering to those are just out of scope, we do not even try).
2. Clear rules about what is our domain to teach and what we can expect customers to find documentation for elsewhere. Example of guidelines:
    - We do not teach people how to program
    - We do not teach people how to debug
    - We do not teach people how to compile and run software in other tools than CMake and Visual Studio
    - We teach how to best integrate Zivid SW with 3rd party. For example Eigen, but we assume the customer knows Eigen as a standalone library, we do not teach Eigen. Then the customer should look up Egen doc on his own.
    - We do not teach best practices in C++ (We do not write an article in KB about RAII), but our examples are written using best practice

Having those two guidelines could help to make documentation more consistent regardless of author and it could help making decisions like this. 

FYI: @ajlandau, @torbsorb, @SatjaSivcev, @runenordmo, @dragly, @Martiix: What do you think about this?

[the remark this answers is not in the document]
nedrebo: > Can you explain what is the benefit of the previous solution, for developers? I don't get it, except that it has 4 lines of code less?

I am not concerned about those four lines in particular, but I am concerned if our default is to inflate documentation by 30% to cater for a marginal or made up customer that will never be able to integrate Zivid into anything at all. All that extra nice is paid for by more experienced developers as mental overhead. My concern is that we optimize for the wrong end-user or that we try to cater to all which is just going to be bad for all.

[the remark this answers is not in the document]
runenordmo: > Samples should be simple. "Look here only 20 lines to get going" and noise like this should be minimized imo.

I completely agree on this, and this was why I added the original comment.

> Having those two guidelines could help to make documentation more consistent regardless of author and it could help making decisions like this.

I definitely think that such guidelines could be helpful, both for writing and reviewing samples and documentation; as this is an example presenting potential misalignment between for instance the software and application team.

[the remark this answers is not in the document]
dragly: I completely understand the purpose of those lines since I also am someone who has double clicked plenty of .exes that immediately closed in sample and bin folders of random software on Windows :facepalm:

I would, however, argue for removing those lines.

Defining the target user is definitely at the core of answering this and similar questions. And defining the target user also involves *choosing* a target user. That choice should be driven by value for our company. For instance, targeting experienced software engineers is likely more valuable than targeting first-year students. 

Now, I am not saying experienced software engineers will not double-click these files (I do it myself), but I assume that most *target users for the samples* are experienced software engineers who will conclude that they clicked a CLI program that needs to opened from CMD. That is, I think the intersection between people who do not understand the problem and are also target users are very few. This might be a wrong assumption, however, so please feel free to correct me on this.

However, if there are real and important users that are dissuaded by these samples not staying opened, perhaps you could you say more about who they are? It sounds like they could be equally important users (and decision makers) who would be much more happy if they found some nice GUI examples when double-clicking these .exes. In that case, we could perhaps divide our samples folder into a root folder with GUI examples and a sub-folder clearly marked as CLI samples?

All in all, I am not too bothered by these lines, but I would keep them out if it was up to me. It sounds like they are the wrong solution for the user who has the CMD closing-problem.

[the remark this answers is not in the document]
ajlandau: I agree that we should set who we expect to be using the samples in the KB somewhere @nedrebo. My mindset when I was adding this is I was following the tutorials and realized with how my first impression is of the samples, that if I have a sample, I would expect it to work with just a double click, and if it doesn't, that it would tell me why. I did not see the few extra lines as a deterrent and if someone does see that, then wouldn't they already not enjoy the fact it is even in a try-catch loop?

If we say that the first person to interact with the camera is an engineer who is just buying one to investigate if they think it's good enough, at least in my experience it is not the guarantee that the person who will be integrating the camera will be the same person to first evaluate it. It is also not a guarantee that this person will be as immediately savvy as one would hope, but probably would have the engineering experience or seniority to still make a call over if it was worth giving to the people that do have the knowledge to integrate it into their systems.

So that boils down to for me is: 
-what do we expect users to gain from these samples?
-Are they for learning how our code works?
- are they for showing off how our camera works in their settings and without the GUI?
-Should we say that the Zivid studio interface is enough of a tool for those who are not SE to show off the camera features?

PR #132, opened 2020-07-03
Update for new copyPointZ API

chrisasc:


SatjaSivcev:
Did you run the CI locally @chrisasc?

source/Applications/Advanced/CreateDepthMap/CreateDepthMap.cpp
```diff
@@ -15,32 +15,26 @@ This example shows how to convert point cloud from ZDF file to OpenCV format, th
...
-        y,
-        z
-    };
-
-    template<Axis axis>
-    float getValue(const Zivid::PointXYZ &p);
-
-    template<>
-    float getValue<Axis::z>(const Zivid::PointXYZ &p)
+    float getValue(const Zivid::PointZ &p)
```
SatjaSivcev, line 18: This is now specific to Z so I would call it getValueZ

source/Applications/Advanced/CreateDepthMap/CreateDepthMap.cpp
```diff
@@ -15,32 +15,26 @@ This example shows how to convert point cloud from ZDF file to OpenCV format, th
...
 
-    template<Axis axis>
-    bool isLesserOrNan(const Zivid::PointXYZ &a, const Zivid::PointXYZ &b)
+    bool isLesserOrNan(const Zivid::PointZ &a, const Zivid::PointZ &b)
     {
-        return getValue<axis>(a) < getValue<axis>(b) ? true : std::isnan(getValue<axis>(a));
+        if(std::isnan(getValue(a)) && std::isnan(getValue(b)))
+        {
+            return false;
+        }
```
SatjaSivcev, line 28: What case do you cover by adding this?

chrisasc: Some compilers dont like to compare a NaN value against another NaN value. This was the error Torbjørn and I got when we were running the sample in debug mode. This little if statement solves it

source/Applications/Advanced/CreateDepthMap/CreateDepthMap.cpp
chrisasc: done

chrisasc:
> Did you run the CI locally @chrisasc?

No, i only committed the PR, Torbjørn wrote the code an ran it.

PR #135, opened 2020-07-06
Add PointCloudTutorial.md

SatjaSivcev:


source/Applications/PointCloudTutorial.md
```diff
@@ -0,0 +1,131 @@
...
+The ```Zivid::Frame``` can be captured with a camera ([go to source][frame-capture]).
+```cpp
+const auto frame = camera.capture(settings);
+```
+Check [Capture Tutorial][capture-tutorial] for more details.
+
+### Load
+It can also be loaded from a ZDF file ([go to source][frame-from-file]).
+```cpp
+const Zivid::Frame frame = Zivid::Frame(Zivid3D.zdf);
```
eskaur, line 32: Should be a string, like "Zivid3D.zdf".
Also consider making it clear that it's a path, like "/path/to/Zivid3D.zdf".

source/Applications/PointCloudTutorial.md
```diff
@@ -0,0 +1,131 @@
...
+
+[//]: ### "Recommended further reading"
+
+[installation-instructions-url]: ../../README.md#instructions
+[frame-from-file]:Basic/FileFormats/ReadIterateZDF/ReadIterateZDF.cpp#L15-L17
+[frame-capture]:../Camera/Basic/Capture/Capture.cpp#L28
+[capture-tutorial]:../Camera/Basic/CaptureTutorial.md#L158
+[point-cloud]:Advanced/Downsample/Downsample.cpp#L181
+[transform]:Applications/Advanced/MultiCamera/StitchByTransformationFromZDF/StitchByTransformationFromZDF.cpp
+[visualize-point-cloud]:Advanced/Downsample/Downsample.cpp#L158-L166
```
eskaur, line 129: Seem strange to use the Downsample sample as the example here. Is there something appropriate in "source/Applications/Basic/Visualization" that could be linked to instead?

source/Applications/PointCloudTutorial.md
```diff
@@ -0,0 +1,131 @@
...
+visualizer.showMaximized();
+visualizer.show(frame);
+visualizer.resetToFit();
+
+visualizer.run();
+```
+
+## Point Cloud
+
+### Get
```
eskaur, line 50: Maybe call this "Get handle", since "Get" sounds like you may be copying.

source/Applications/PointCloudTutorial.md
```diff
@@ -0,0 +1,131 @@
...
+
+You can now get a handle to the point cloud data on the GPU ([go to source][point-cloud]).
+```cpp
+frame.pointCloud();
+```
+Note: ```Zivid::Camera::capture()``` method returns as soon as the camera is done capturing. However, the point cloud becomes available only after the GPU is finished processing. Therefore, if you call ```Zivid::Frame::pointCloud()``` right after the capture, the CPU will wait until the point cloud is available and then return.
+
```
eskaur, line 57: The first sentence is accurate. However, the rest is inaccurate, probably because I said something wrong to you earlier.

`Frame::pointCloud()` always returns instantly. Any wait-time for processing occurs when you start calling `pointCloud.copyXXXX()` functions. Perhaps it can be something like this:

"
Note: ```Zivid::Camera::capture()``` method returns as soon as the camera is done capturing. At that point the handle from `frame.pointCloud()` is available instantly as well. However, the actual point cloud data becomes available only after the GPU is finished processing. This processing will occur on the GPU in the background, and any calls to data-copy functions (see section "Copy" below) will wait for this to finish before proceeding with the requested copy operation.
"

source/Applications/PointCloudTutorial.md
```diff
@@ -0,0 +1,131 @@
...
+This call does not perform any copying from GPU memory.
```
eskaur, line 58: Maybe be even more precise: "The call to `frame.pointCloud()` does not perform any copying from GPU memory."

source/Applications/PointCloudTutorial.md
```diff
@@ -0,0 +1,131 @@
...
+```cpp
+const auto pointCloud = frame.pointCloud(); /* Get a handle to the full point cloud on the GPU. */
+
+auto rgba = cv::Mat(pointCloud.height(), pointCloud.width(), CV_8UC4); /* Allocate an OpenCV matrix with
+an appropriate size. */
+
+auto *dataPtr = reinterpret_cast<Zivid::ColorRGBA *>(rgba.data); /* Cast the OpenCV data pointer to
+ColorRGBA* so the Zivid API can understand which data to copy. */
+
+pointCloud.copyData(dataPtr); /* Copy RGBA data straight into the OpenCV memory buffer. */
```
eskaur, line 106: Maybe let the comment be "Copy RGBA data straight from GPU memory into the OpenCV memory buffer."
More accurate, and also sounds more impressive ;)

source/Applications/PointCloudTutorial.md
```diff
@@ -0,0 +1,131 @@
...
+This tutorial shows how to use the Zivid SDK to extract the point cloud, manipulate it, transform it, and visualize it.
+
+[//]: ### "Recommended further reading"
+
+[installation-instructions-url]: ../../README.md#instructions
+[frame-from-file]:Basic/FileFormats/ReadIterateZDF/ReadIterateZDF.cpp#L15-L17
+[frame-capture]:../Camera/Basic/Capture/Capture.cpp#L28
+[capture-tutorial]:../Camera/Basic/CaptureTutorial.md#L158
+[point-cloud]:Advanced/Downsample/Downsample.cpp#L181
+[transform]:Applications/Advanced/MultiCamera/StitchByTransformationFromZDF/StitchByTransformationFromZDF.cpp
```
eskaur: This link doesn't work, because it gets "Applications" twice in the path. I think this is because the paths are relative to the location of this file (notice how you don't write "Applications/" in the other links here.

eskaur:
First review done, @SatjaSivcev

source/Applications/PointCloudTutorial.md
eskaur: Ah you fixed this in the meantime :)

SatjaSivcev: yes, done :)

source/Applications/PointCloudTutorial.md
SatjaSivcev: done

source/Applications/PointCloudTutorial.md
SatjaSivcev: done using captureVis3D

source/Applications/PointCloudTutorial.md
SatjaSivcev: done

source/Applications/PointCloudTutorial.md
SatjaSivcev: Done, thanks for the suggestion. Used it.

source/Applications/PointCloudTutorial.md
SatjaSivcev: Done, but usig this: The method ```Zivid::Frame.pointCloud()``` does not perform any copying from GPU memory.
Is it correct?

source/Applications/PointCloudTutorial.md
SatjaSivcev: done

SatjaSivcev:
Ready for the next review

SatjaSivcev:
Thanks and thank you for reviewing in a short period.

Issue #98, opened 2020-04-28
Print 'Sigma' for ZividBenchmark


SatjaSivcev:
The Gaussian filter time depends on sigma. Therefore, the tool should at least print which one it is using.

PR #139, opened 2020-07-20
Print test headers in ZividBenchmark.cpp

### LINKED ISSUE #131: Add test names for each test in ZividBenchmark
This will make it easier to compare results from multiple cameras, e.g. Test1 Test2 would work.

SatjaSivcev:
Fixes #131 and #98

### DIFF
--- a/source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
+++ b/source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
@@ -67,11 +67,15 @@ namespace
     {
         if(settings.processing().filters().smoothing().gaussian().isEnabled().value())
         {
+            std::string gaussianString;
+            gaussianString = std::string{ "Gaussian (Sigma = " }
+                             + settings.processing().filters().smoothing().gaussian().sigma().toString() + " )";
+
             if(settings.processing().filters().reflection().removal().isEnabled().value())
             {
-                return "{ Gaussian, Reflection }";
+                return "{ " + gaussianString + ", Reflection }";
             }
-            return "{ Gaussian }";
+            return "{ " + gaussianString + " }";
         }
         if(settings.processing().filters().reflection().removal().isEnabled().value())
         {
@@ -107,6 +111,14 @@ namespace
                   << stringList.at(1) << stringList.at(2) << std::endl;
     }
 
+    void printHeader(const std::string &firstString)
+    {
+        printPrimarySeparationLine();
+        std::cout << std::endl;
+        printPrimarySeparationLine();
+        std::cout << firstString << std::endl;
+    }
+
     void printHeaderLine(const std::string &firstString, size_t num, const std::string &secondString)
     {
         printPrimarySeparationLine();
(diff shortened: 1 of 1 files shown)


ajlandau:
The changes you made definitely make it better to read. It adds clarity to what is happening at each step.

I do have one question. When we switch to SDK 2.0 we also switch how we define the aperture as the f-stop and now use decimals, where before it was always whole numbers. Now that we use decimates, should we truncate to after the first two places? And should we do the same with the Sigma value? I think it would just look cleaner if we do. The image shows your code now to left and then the 1.8.1 benchmark run on my computer to the right for reference for the previous view to the right side. 

image

SatjaSivcev:
> 
> 
> The changes you made definitely make it better to read. It adds clarity to what is happening at each step.
> 
> I do have one question. When we switch to SDK 2.0 we also switch how we define the aperture as the f-stop and now use decimals, where before it was always whole numbers. Now that we use decimals, should we truncate to after the first two places? And should we do the same with the Sigma value? I think it would just look cleaner if we do. The image shows your code now to left and then the 1.8.1 zividbenchmark run on my computer to the right for reference for the previous view.
> 
> image

Thanks @ajlandau. Very good that you've attached the screenshot. I should have done that in the first place. I agree that it would look nicer if we were to truncate it. I wanted to do it but it seemed verbose to do it in C++, I will explore a bit more.

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
```diff
@@ -118,6 +130,12 @@ namespace
         printHeaderLine("Connecting and disconnecting ", numConnects, " times each (be patient):");
     }
 
+    void printCapture3DAndFiltersHeader(const std::string &subtest)
```
torbsorb, line 133: From the arguments it looks like this function should be called `printSubtest`, or something like that.

torbsorb:
> 
> 
> > The changes you made definitely make it better to read. It adds clarity to what is happening at each step.
> > I do have one question. When we switch to SDK 2.0 we also switch how we define the aperture as the f-stop and now use decimals, where before it was always whole numbers. Now that we use decimals, should we truncate to after the first two places? And should we do the same with the Sigma value? I think it would just look cleaner if we do. The image shows your code now to left and then the 1.8.1 zividbenchmark run on my computer to the right for reference for the previous view.
> > image
> 
> Thanks @ajlandau. Very good that you've attached the screenshot. I should have done that in the first place. I agree that it would look nicer if we were to truncate it. I wanted to do it but it seemed verbose to do it in C++, I will explore a bit more.
> 
> The problem with the current implementation is that both aperture and sigma values come as strings so we would need a function to reduce the num of dec places. Alternatively, add a function for rounding somewhere else.

http://www.cplusplus.com/reference/ios/ios_base/precision/

source/Camera/InfoUtilOther/ZividBenchmark/ZividBenchmark.cpp
SatjaSivcev: Done - printSubtestHeader

PR #140, opened 2020-07-29
Improve depth map visualization in CreateDepthMap.cpp

### LINKED ISSUE #138: Improve depth visualization in CreateDepthMap
From @dragly :
_Use a different colormap - jet is bad for a number of reasons. I think we should
use Virids for things like depth because it is going from dark to light. If we really
want rainbows, consider using Turbo. But I do not think there is a good reason to pick
a diverging colormap like jet or turbo for something like depth._
```
cv::Mat zJetColorMap;
cv::applyColorMap(z, zJetColorMap, cv::COLORMAP_JET);
```

SatjaSivcev:
Jet colormap is replaced by Viridis because it gives a better perception of depth.

Fixes #138

### DIFF
--- a/source/Applications/Advanced/CreateDepthMap/CreateDepthMap.cpp
+++ b/source/Applications/Advanced/CreateDepthMap/CreateDepthMap.cpp
@@ -79,8 +79,8 @@ namespace
         }
 
         // Applying color map
-        cv::Mat zJetColorMap;
-        cv::applyColorMap(z, zJetColorMap, cv::COLORMAP_JET);
+        cv::Mat zColorMap;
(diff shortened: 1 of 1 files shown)


SatjaSivcev:
@dragly can you please review?

dragly:
Oooh, nice! I will 👍

SatjaSivcev:
> 
> 
> Oooh, nice! I will 👍

Any chance you can do this?

dragly (approved):
Sorry about the very long time to get back to you. This totally slipped my mind and got lost in other notifications.

The results look very good!

PR #144, opened 2020-09-08
Update to Zivid SDK v2.1.0 and modify downsample sample

robertkt1:
Almost zero code to write. Deleted a lot of code and changed a couple of lines to match built-in function

### DIFF
--- a/source/Applications/Advanced/Downsample/Downsample.cpp
+++ b/source/Applications/Advanced/Downsample/Downsample.cpp
@@ -6,154 +6,10 @@ main instructions for Zivid samples.
 #include <Zivid/Visualization/Visualizer.h>
 #include <Zivid/Zivid.h>
 
-#include <Eigen/Core>
-#include <Eigen/Dense>
-#include <pcl/io/pcd_io.h>
-#include <pcl/io/ply_io.h>
-#include <pcl/visualization/cloud_viewer.h>
-
 #include <iostream>
 
 namespace
 {
-    float nanToZero(float x)
-    {
-        if(std::isnan(x))
-        {
-            return 0;
-        }
-        return x;
-    }
-
-    Eigen::MatrixXf lineSum(Eigen::MatrixXf matrixf, int downsamplingFactor)
-    {
-        Eigen::Map<Eigen::MatrixXf> flattenedMatrixMap(matrixf.data(),
-                                                       downsamplingFactor,
-                                                       matrixf.rows() * matrixf.cols() / downsamplingFactor);
-        std::function<float(float)> nanToZeroFunctor = nanToZero;
-
-        Eigen::MatrixXf flattenedMatrixNansRemoved = flattenedMatrixMap.unaryExpr(nanToZeroFunctor);
-
-        Eigen::MatrixXf flattenedMatrixColwiseSum = flattenedMatrixNansRemoved.colwise().sum();
-
-        Eigen::Map<Eigen::MatrixXf> reshapedMatrixMap(flattenedMatrixColwiseSum.data(),
-                                                      matrixf.rows() / downsamplingFactor,
-                                                      matrixf.cols());
-
-        return reshapedMatrixMap;
-    }
-
-    Eigen::MatrixXf gridSum(const Eigen::MatrixXf &matrixf, int downsamplingFactor)
-    {
-        return lineSum(lineSum(matrixf, downsamplingFactor).transpose(), downsamplingFactor).transpose();
-    }
-
-    Eigen::MatrixXi downsampleAndRound(const Eigen::MatrixXi &matrixi, int downsamplingFactor)
-    {
-        Eigen::MatrixXf matrixf = matrixi.template cast<float>();
-
-        std::function<float(float)> roundFunctor = [](float f) { return std::round(f); };
-
-        return ((gridSum(matrixf, downsamplingFactor) / (downsamplingFactor * downsamplingFactor))
-                    .unaryExpr(roundFunctor))
-            .template cast<int>();
-    }
-
-    pcl::PointCloud<pcl::PointXYZRGB> downsample(const Zivid::PointCloud &pointCloud, int downsamplingFactor)
-    {
-        /*
-        Function for downsampling a Zivid point cloud. The downsampling factor represents the denominator
-        of a fraction that represents the size of the downsampled point cloud relative to the original
-        point cloud, e.g. 2 - one-half,  3 - one-third, 4 one-quarter, etc.
-        */
-
-        if((pointCloud.height() % downsamplingFactor) != 0U || (pointCloud.width() % downsamplingFactor) != 0U)
-        {
-            throw std::invalid_argument("Downsampling factor (" + std::to_string(downsamplingFactor)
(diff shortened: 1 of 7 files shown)


source/Applications/Advanced/Downsample/Downsample.cpp
```diff
@@ -182,16 +38,14 @@ int main()
         const auto pointCloud = frame.pointCloud();
         std::cout << "Loaded " << pointCloud.size() << " data points" << std::endl;
 
-        auto downsamplingFactor = 4;
-        auto pointCloudDownsampled = downsample(pointCloud, downsamplingFactor);
+        const auto downsamplingFactor = Zivid::PointCloud::Downsampling::by2x2;
```
SatjaSivcev, line 41: Add the following line before
`std::cout << "Downsampling point cloud" << std::endl;`

source/Applications/Advanced/Downsample/Downsample.cpp
```diff
@@ -182,16 +38,14 @@ int main()
...
+        const auto downsampledPointCloud = pointCloud.downsampled(downsamplingFactor);
```
SatjaSivcev, line 42: One liner? No need for downsamplingFactor variable I think

SatjaSivcev (changes requested):
Minor changes required.
Have you tested that this works ok?
P.S. You will need one PR before this one where you modify:
- CMakeLists - set(ZIVID_VERSION 2.0.0)
- CI files

SatjaSivcev:
In Readme.md you have 2.0.0 hardcoded. Fix to 2.1.0

Tip: Open the whole project in Visual Studio Code and search for 2.0.

SatjaSivcev (commented):
One last comment to fix. We can wait until the release candidate now.

SatjaSivcev:
> 
> 
> One last comment to fix. We can wait until the release candidate now.

This is done. Waiting for the release now.

SatjaSivcev:
It seems this sample is using 'fileName ' and other ones that load zdf use 'dataFile'. Can you check and confirm and if true modifiy this one?

SatjaSivcev:
You can rebase on master and squash commits. Perhaps it is best to have a single commit message called stg like 'Update samples for SDK 2.1.0' (check commit history to see what we did), and in that commit you can specify what has been done (This commit modified Downsample sample, etc...

source/Applications/Advanced/Downsample/Downsample.cpp
```diff
@@ -175,23 +31,24 @@ int main()
...
+        std::string dataFile = std::string(ZIVID_SAMPLE_DATA_DIR) + "/Zivid3D.zdf";
+        std::cout << "Reading ZDF frame from file: " << dataFile << std::endl;
+        const auto frame = Zivid::Frame(dataFile);
 
         const auto pointCloud = frame.pointCloud();
-        std::cout << "Loaded " << pointCloud.size() << " data points" << std::endl;
+        std::cout << "Size of point cloud before downsampling: " << pointCloud.size() << " data points" << std::endl;
+
+        std::cout << "Downsampling point cloud" << std::endl;
+        const auto downsampledPointCloud = pointCloud.downsampled(Zivid::PointCloud::Downsampling::by2x2);
```
runenordmo, line 42: We also support in-place downsampling; `downsample()`. I think that is more relevant for the sample, since that is more performant and should be the default.

An good alternative is to show both in-place downsampling (`downsample()`) and downsampling into a new `Pointcloud` instance (`downsampled()`).

runenordmo:
LGTM, will likely approve when my comment on doing the downsampling in-place is addressed.

source/Applications/Advanced/Downsample/Downsample.cpp
eskaur: Do you suggest moving the first visualizer call to before "downsample" then?

robertkt1: That's how I solved it now. By moving the first viz

runenordmo (approved):
LGTM :+1: Can you just rebase and squash the two commits into one?
You can also test that the sample still works (win/ubuntu)

runenordmo:
@robertkt1, there is a spelling error in the commit message: 
`Update downsamnple sample for SDK 2.1`

runenordmo (changes requested):
You need to correct the commit message spelling error:
`Update downsamnple sample for SDK 2.1`

source/Applications/Advanced/Downsample/Downsample.cpp
SatjaSivcev: > 
> 
> We also support in-place downsampling; `downsample()`. I think we should rather use that in this sample, since it is more performant (no copy) and should definitely be the default. @robertkt1
> 
> An alternative is to show both in-place downsampling (`downsample()`) and downsampling into a new `Pointcloud` instance (`downsampled()`).

@robertkt1 Rune's comment on two options should be covered in the point cloud tutorial: https://github.com/zivid/zivid-cpp-samples/blob/new-downsample-sample/source/Applications/PointCloudTutorial.md

runenordmo:
You should format your code in `Downsample.cpp`, @robertkt1. Run `clang-format -i -style=file <path_to>/Downsample.cpp`

source/Applications/Advanced/Downsample/Downsample.cpp
```diff
@@ -175,23 +31,23 @@ int main()
...
-        const auto pointCloud = frame.pointCloud();
-        std::cout << "Loaded " << pointCloud.size() << " data points" << std::endl;
+        std::string dataFile = std::string(ZIVID_SAMPLE_DATA_DIR) + "/Zivid3D.zdf";
+        std::cout << "Reading ZDF frame from file: " << dataFile << std::endl;
+        const auto frame = Zivid::Frame(dataFile);
 
-        auto downsamplingFactor = 4;
-        auto pointCloudDownsampled = downsample(pointCloud, downsamplingFactor);
+        auto pointCloud = frame.pointCloud();
+        std::cout << "Size of point cloud before downsampling: " << pointCloud.size() << " data points" << std::endl;
```
runenordmo, line 39: Nitpick:
The output of this sample is now:
```
1. Size of point cloud before downsampling: 2861568 data points
2. Visualizing point cloud before downsampling
3. Setting up visualization
4. Visualizing point cloud
5. Running visualizer. Blocking until window closes
6. Downsampling point cloud
7. Size of point cloud after downsampling: 178848 data points
8. Visualizing point cloud after downsampling
9. Setting up visualization
10. Visualizing point cloud
11. Running visualizer. Blocking until window closes
```
**Line 2-5 and 8-11 is basically the same information given four lines in a row.
I think you can remove two of the printouts from inside `visualizePointCloud()`,** given that is does not give any inconsistency with the other samples.

runenordmo (commented):
This looks close to merge to me now :+1:  I added only one nitpick comment. When you answer that, I'll approve.

Btw, have you tested running the samples for 2.1 (should be no reason they should not)? It looks like the CI pipeline only test that the samples build.

runenordmo (approved):
Nice. Now you can squash your fixup commit into the original commit, then rebase on master and merge when CI's green :+1:

SatjaSivcev (approved):

PR #147, opened 2020-11-11
Remove the need to enter \n to return EXIT_FAILURE

robertkt1:
Change exception handling to always return EXIT_FAILURE, when handling exception. 
Before EXIT_FAILURE was only returned if one enters '\n' after an exception catch


SatjaSivcev:
Change `std::cout << "Press enter to exit." << std::endl;` to `std::cout << "Press a key to exit." << std::endl;`

SatjaSivcev (changes requested):
Minor change.

SatjaSivcev:
> 
> 
> Change `std::cout << "Press enter to exit." << std::endl;` to `std::cout << "Press a key to exit." << std::endl;`

Not needed since it will exit on pressing enter with no character input.

SatjaSivcev (approved):

PR #153, opened 2021-01-08
Sample: Add TransformViaArucoMarker sample

chrisasc:
The sample reads a point cloud, detects an aruco marker and its pose.
Then transforms to this frame and saves the point cloud.


chrisasc:
I need some help with understanding the CI linting errors...

README.md
```diff
@@ -55,6 +55,10 @@ There are two main categories of samples: **Camera** and **Applications**. The s
     - [**MultiCamera**][MultiCamera-url] ([tutorial][MultiCameraTutorial-url])
       - [**MultiCameraCalibration**][MultiCameraCalibration-url] and [**MultiCameraCalibrationFromFile**][MultiCameraCalibrationFromZDF-url] - Use captures of a calibration object to generate transformation matrices to a single coordinate frame. First sample captures from connected cameras, the second loads existing captures from .ZDF files.
       - [**StitchByTransformation**][StitchByTransformation-url] and [**StitchByTransformationFromFile**][StitchByTransformationFromZDF-url] - Use transformation matrices from Multi-Camera calibration to transform point clouds into single coordinate frame. First sample uses connected cameras, the second loads existing captures from .ZDF files.
+    - [**TransformViaArucoMarker**][TransformViaArucoMarker-url] - Transform a point cloud from camera frame to an aruco marker frame using the markers pose.
```
SatjaSivcev, line 58: 'marker pose' or 'marker's pose'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+    {
+        Zivid::Application zivid;
+
+        const auto dataFile = std::string(ZIVID_SAMPLE_DATA_DIR) + "/ArucoSample.zdf";
+        std::cout << "Reading ZDF frame from file: " << dataFile << std::endl;
+        const Zivid::Frame frame = Zivid::Frame(dataFile);
+        auto pointCloud = frame.pointCloud();
+
+        std::cout << "Converting from RGB to OpenCV BGR image" << std::endl;
+        cv::Mat image = pointCloudToBGR(pointCloud);
```
SatjaSivcev, line 275: 'bgr' instead of 'image'? Isn't that what we use across samples?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+        {
+            marker.draw(image);
+        }
+        std::cout << "Displaying detected markers" << std::endl;
+        cv::imshow("Detected Aruco markers", image);
+        cv::waitKey(0);
+
+        std::cout << "Detecting aruco marker poses" << std::endl;
+        const auto display = true;
+        auto transform = getMarkerPose(pointCloud, markers[0], display);
```
SatjaSivcev, line 291: why is this a variable? Why not use it only in line 291?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+            }
+        }
+        for(int i = 0; i < 3; ++i)
+        {
+            cvInverted.at<float>(i, 3) = newTrans.at<float>(i, 0);
+        }
+        cvInverted.at<float>(3, 3) = 1.0;
+
+        auto *ptr = &cvInverted.at<float>(0, 0);
+        auto invertTransform = Zivid::Matrix4x4(ptr, ptr + 16);
```
SatjaSivcev, line 257: call this 'invertedTransform'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+        cv::waitKey(0);
+
+        std::cout << "Detecting aruco marker poses" << std::endl;
+        const auto display = true;
+        auto transform = getMarkerPose(pointCloud, markers[0], display);
+
+        std::cout << "Aruco marker transformation matrix:" << std::endl;
+        std::cout << transform << std::endl;
+
+        auto transformInv = invertAffineTransformation(transform);
```
SatjaSivcev, line 296: call this invertedTransform?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+        const auto centers3D = estimate3DMarkerPoints(pointCloud, centers2D);
+
+        auto origin = cv::Vec3d{ centers3D[0].x, centers3D[0].y, centers3D[0].z };
+        auto yAxis = cv::Vec3d{ corners3D[0].x - corners3D[1].x,
+                                corners3D[0].y - corners3D[1].y,
+                                corners3D[0].z - corners3D[1].z };
+
+        auto xAxis = cv::Vec3d{ corners3D[2].x - corners3D[1].x,
+                                corners3D[2].y - corners3D[1].y,
+                                corners3D[2].z - corners3D[1].z };
```
SatjaSivcev, line 197: Why yAxis first?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+        cv::imshow("Aruco marker pose", bgr);
+        cv::waitKey(0);
+    }
+
+    Zivid::Matrix4x4 getMarkerPose(const Zivid::PointCloud &pointCloud,
+                                   const aruconano::Marker &markers,
+                                   const bool display = false)
+    {
+        const auto corners2D = get2DMarkerCorners(markers);
```
SatjaSivcev, line 184: Add a comment sentence about what is being done in this function and how, if possible?
Or add a number of smaller comments from line 190 to line 212 for each operation?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+        auto xAxis = cv::Vec3d{ corners3D[2].x - corners3D[1].x,
+                                corners3D[2].y - corners3D[1].y,
+                                corners3D[2].z - corners3D[1].z };
+
+        auto u = xAxis / cv::norm(xAxis, cv::NORM_L2);
+        auto v = yAxis / cv::norm(yAxis, cv::NORM_L2);
+        auto normal = u.cross(v);
+        auto unitNormal = normal / cv::norm(normal, cv::NORM_L2);
+
+        auto rotMatrix = cv::Mat(3, 3, CV_64F);
```
SatjaSivcev, line 204: use 'rotationMatrix' - we tend to use fullNames for variables when we can. Try to implement this for the whole sample.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+        for(int i = 0; i < 3; ++i)
+        {
+            rotMatrix.at<double>(i, 0) = u[i];
+            rotMatrix.at<double>(i, 1) = v[i];
+            rotMatrix.at<double>(i, 2) = unitNormal[i];
+        }
+
+        auto transform = transformationMatrix(rotMatrix, origin);
+
+        if(display)
```
SatjaSivcev, line 214: Why make this an option? Why not show it every time?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+    void displayMarkerPose(const Zivid::PointCloud &pointCloud,
+                           const std::vector<cv::Point2f> &corners2D,
+                           const std::vector<cv::Point2f> &center2D)
+    {
+        auto rgbaImage = cv::Mat(pointCloud.height(), pointCloud.width(), CV_8UC4); // NOLINT(hicpp-signed-bitwise)
+        auto *ptr =
+            reinterpret_cast<Zivid::ColorRGBA *>(rgbaImage.data); // NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
+        pointCloud.copyData(ptr);
+        auto bgr = cv::Mat(pointCloud.height(), pointCloud.width(), CV_8UC4); // NOLINT(hicpp-signed-bitwise)
+        cv::cvtColor(rgbaImage, bgr, cv::COLOR_BGR2RGB);
```
SatjaSivcev, line 166: You have a function pointCloudToBGR(), why not use it?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+        std::vector<cv::Point2f> markerCenters2D;
+        markerCenters2D.reserve(markers.size());
+
+        auto p11 = markers[0];
+        auto p12 = markers[1];
+        auto p21 = markers[3];
+        auto p22 = markers[2];
+
+        const auto backDiagonal{ fitLine(p11, p22) };
+        const auto forwardDiagonal{ fitLine(p21, p12) };
```
SatjaSivcev, line 66: What made you use these variable names? They are not intuitive to me.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+    std::vector<cv::Point3d> estimate3DMarkerPoints(const Zivid::PointCloud &pointCloud,
+                                                    const std::vector<cv::Point2f> &markerPoints2D)
+    {
+        if(markerPoints2D.empty())
+        {
+            return {};
+        }
+
+        int w = pointCloud.width();
+        int h = pointCloud.height();
```
SatjaSivcev, line 103: 'w' to 'width', 'h' to 'height'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+        markerPoints3D.reserve(markerPoints2D.size());
+        for(const auto &point2D : markerPoints2D)
+        {
+            float x = point2D.x;
+            float y = point2D.y;
+
+            int x1 = clamp(static_cast<int>(std::floor(x)), 0, w - 1);
+            int x2 = clamp(static_cast<int>(std::ceil(x)), 0, w - 1);
+            int y1 = clamp(static_cast<int>(std::floor(y)), 0, h - 1);
+            int y2 = clamp(static_cast<int>(std::ceil(y)), 0, h - 1);
```
SatjaSivcev, line 116: What does this do? Add a comment or rename variables so that it is more clear?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+            auto q11 = points(y1, x1);
+            auto q12 = points(y2, x1);
+            auto q21 = points(y1, x2);
+            auto q22 = points(y2, x2);
```
SatjaSivcev, line 121: Why use 'q'? Is it a marker corner? If so, use markerCorner in the var name.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+            float a = (x2 - x) / (x2 - x1); // NOLINT(cppcoreguidelines-narrowing-conversions)
+            float b = (x - x1) / (x2 - x1); // NOLINT(cppcoreguidelines-narrowing-conversions)
+
+            const cv::Point3f fxy1{ a * q11.x + b * q21.x, a * q11.y + b * q21.y, a * q11.z + b * q21.z };
+            const cv::Point3f fxy2{ a * q12.x + b * q22.x, a * q12.y + b * q22.y, a * q12.z + b * q22.z };
```
SatjaSivcev, line 127: fxy1 and fxy2 are bad variable names

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+            int y1 = clamp(static_cast<int>(std::floor(y)), 0, h - 1);
+            int y2 = clamp(static_cast<int>(std::ceil(y)), 0, h - 1);
+
+            auto q11 = points(y1, x1);
+            auto q12 = points(y2, x1);
+            auto q21 = points(y1, x2);
+            auto q22 = points(y2, x2);
+
+            float a = (x2 - x) / (x2 - x1); // NOLINT(cppcoreguidelines-narrowing-conversions)
+            float b = (x - x1) / (x2 - x1); // NOLINT(cppcoreguidelines-narrowing-conversions)
```
SatjaSivcev, line 124: What are a and b? I don't like the variable names

SatjaSivcev: What do these lines of code do? Think about adding comments on grouped operations where needed.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+            const cv::Point3f fxy1{ a * q11.x + b * q21.x, a * q11.y + b * q21.y, a * q11.z + b * q21.z };
+            const cv::Point3f fxy2{ a * q12.x + b * q22.x, a * q12.y + b * q22.y, a * q12.z + b * q22.z };
+
+            float c = (y2 - y) / (y2 - y1); // NOLINT(cppcoreguidelines-narrowing-conversions)
+            float d = (y - y1) / (y2 - y1); // NOLINT(cppcoreguidelines-narrowing-conversions)
```
SatjaSivcev, line 130: Same here, see above.

SatjaSivcev (changes requested):
Please address the comments.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: We actually use a few different things.
CaptureUndistortRGB -> cv::Mat imageToBGR()
CreateDepthMap and MaskPointCloud -> pointCloudToCvZ() and pointCloudToCvBGR()

I would either rename it to
        std::cout << "Converting to BGR image" << std::endl;
        cv::Mat image = pointCloudToCvBGR(pointCloud);
or
        std::cout << "Converting to OpenCV BGR image format" << std::endl;
        cv::Mat image = pointCloudToImage(pointCloud);

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: no reason, and it the order does not matter. I can switch the order as its more common to do x first and then y

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: I am not talking about the comment, but about the variable 'image'.
https://github.com/zivid/zivid-cpp-samples/blob/4fd1d27cf681f360a5ecd3ee220afc6a1170a1dd/source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp#L67
https://github.com/zivid/zivid-cpp-samples/blob/4fd1d27cf681f360a5ecd3ee220afc6a1170a1dd/source/Applications/Advanced/CreateDepthMap/CreateDepthMap.cpp#L108

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: The functions clamps/clips a number (here x and y) between the interval 0 and w-1 or h-1. If it is within, the output is x, if its below, the output is 0 and if its above the output is w-1 or h-1. 

From C++17 this is an internal function in std, but since we require that the SDK is functional with C++11, I needed to create it myself. https://en.cppreference.com/w/cpp/algorithm/clamp

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: This is the same notation as the Wikipedia article explains, https://en.wikipedia.org/wiki/Bilinear_interpolation. I added a comment that this is what I do and therefore I use q, fxy1, fxy2, a, b, c, and d.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: I understand what the function does. But it is not clear why it is needed in the algorithm. The answer has to do with the bilinear interpolation. Adding comments in the code to explain the algorithm a bit will solve this.

SatjaSivcev: And by changing the variable names I meant that e.g. xRounded or something like that would say more than x1.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: From wiki: Suppose that we want to find the value of the unknown function f at the point (x, y). It is assumed that we know the value of f at the four points Q11 = (x1, y1), Q12 = (x1, y2), Q21 = (x2, y1), and Q22 = (x2, y2).

Then a comment like 'Getting 2D coordinates for the four points whose 3D coordinates are known' would be helpful perhaps?

README.md
chrisasc: changed to marker's pose

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: no need, I removed the display option.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: removed the option

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: good point. Guess I mixed up what functions I had available then I wrote the displayMarkerPose. Using the pointCloudToBGR() now.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Added a comment explaining this step. But i'm not sure about the best naming for these two...

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: see comment above

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: see comment about bilinear interpolation

README.md
```diff
@@ -56,6 +56,10 @@ There are two main categories of samples: **Camera** and **Applications**. The s
     - [**MultiCamera**][MultiCamera-url] ([tutorial][MultiCameraTutorial-url])
       - [**MultiCameraCalibration**][MultiCameraCalibration-url] and [**MultiCameraCalibrationFromFile**][MultiCameraCalibrationFromZDF-url] - Use captures of a calibration object to generate transformation matrices to a single coordinate frame. First sample captures from connected cameras, the second loads existing captures from .ZDF files.
       - [**StitchByTransformation**][StitchByTransformation-url] and [**StitchByTransformationFromFile**][StitchByTransformationFromZDF-url] - Use transformation matrices from Multi-Camera calibration to transform point clouds into single coordinate frame. First sample uses connected cameras, the second loads existing captures from .ZDF files.
+    - [**TransformViaArucoMarker**][TransformViaArucoMarker-url] - Transform a point cloud from camera frame to an aruco marker frame using the marker's pose.
```
SatjaSivcev, line 59: Transform a point cloud from **camera frame** to **an aruco marker frame** using the marker's pose

On aruco you have indefinite article **an** - get rid of it or add an article to camera frame as well. Compact solution is '..from camera to aruco frame using marker's pose' (which is what I would chose)

README.md
```diff
@@ -56,6 +56,10 @@ There are two main categories of samples: **Camera** and **Applications**. The s
...
+      - **Dependecies:**
+        - [OpenCV](https://opencv.org/) version 4.1.0 or newer
+        - [Aruco Nano](https://www.uco.es/investiga/grupos/ava/node/26) version 3.1.12 or newer. A  header file is included in the sample folder.
```
SatjaSivcev, line 62: ...r. A  header... -double space between 'A' and 'header'

README.md
```diff
@@ -183,3 +187,4 @@ Tip: If your build hangs, try to increase the memory available to Docker.
 [CaptureUndistortRGB-url]: source/Applications/Advanced/CaptureUndistortRGB/CaptureUndistortRGB.cpp
 [CreateDepthMap-url]: source/Applications/Advanced/CreateDepthMap/CreateDepthMap.cpp
 [MaskPointCloud-url]: source/Applications/Advanced/MaskPointCloud/MaskPointCloud.cpp
+[TransformViaArucoMarker-url]: source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```
SatjaSivcev, line 190: TransformViaArucoMarker - I am wondering if we should rename the sample to TransformPointCloud or TransformPointCloudViaArucoMarker.

The reason is that 'transform' on its own can mean anything. The description of the sample is great and clear, but can we do something to improve the sample name as well? We started with 'transform' because that is how our feature is called. Shold we consider 'change of origin' or 'point cloud origin change' or something like that instead? Let's brainstorm a bit during the review.

Whatever you decide don't implement before the review is finished.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
+/*
+This example shows how to transform a point cloud using the pose of an aruco marker.
```
SatjaSivcev, line 2: Use the same description (or more detailed, thus longer) as on line 59.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include <Zivid/Zivid.h>
+
+#include <algorithm>
+#include <opencv2/core/core.hpp>
+#include <opencv2/highgui/highgui.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+
+#include <aruco_nano.h>
+#include <cmath>
+#include <iostream>
```
SatjaSivcev, line 16: why space in between these includes? Can you check our other samples for guidelines on include order? There should be something related to first standard libraries and then 3rd party specialized libraries if I remember correctly. Perhaps CI takes care of that as well, not sure, but check please.

SatjaSivcev:
@chrisasc Have you introduced any // NOLINT that is not already used in the samples? If not sure, check please.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        std::vector<cv::Point2f> markerCenters2D;
+        markerCenters2D.reserve(markers.size());
+
+        const auto corner0 = markers[0];
+        const auto corner1 = markers[1];
+        const auto corner2 = markers[2];
+        const auto corner3 = markers[3];
+
+        // Fit line between two diagonal corners
```
SatjaSivcev, line 66: Use '-ing' - continuous tense for comments.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        auto rgb = cv::Mat(pointCloud.height(), pointCloud.width(), CV_8UC4); // NOLINT(hicpp-signed-bitwise)
+        pointCloud.copyData(
+            reinterpret_cast<Zivid::ColorRGBA *>(rgb.data)); // NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
+        auto bgr = cv::Mat(pointCloud.height(), pointCloud.width(), CV_8UC4); // NOLINT(hicpp-signed-bitwise)
+        cv::cvtColor(rgb, bgr, cv::COLOR_RGBA2BGR);
+
+        return bgr;
+    }
+
+    std::vector<cv::Point2f> estimate2DMarkerCenters(const aruconano::Marker &markers)
```
SatjaSivcev, line 50: Why is 'markers' plural? Isn't it a single marker?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        markerCenters2D.reserve(markers.size());
+
+        const auto corner0 = markers[0];
+        const auto corner1 = markers[1];
+        const auto corner2 = markers[2];
+        const auto corner3 = markers[3];
+
+        // Fit line between two diagonal corners
+        const auto backDiagonal{ fitLine(corner2, corner0) };
+        const auto forwardDiagonal{ fitLine(corner3, corner1) };
```
SatjaSivcev, line 68: Why 'forward' and 'back' diagonal?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: Fitting a line between two diagonal corners of a marker

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        // Find intersection of the two lines
```
SatjaSivcev, line 70: -ing

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        double slope;
+        double intercept;
+    };
+
+    int clamp(const int n, const int lower, const int upper)
+    {
+        return std::max(lower, std::min(n, upper));
+    }
+
+    // Fit a line y=a*x + b to 2 points
```
SatjaSivcev, line 31: In Python samples function headers are inside the function definition, check https://github.com/zivid/zivid-python-samples/blob/6c085911df130417732389d973ec90cded731de7/source/applications/advanced/gamma_correction.py#L28-L50
Please check if this should be the case with C++ as well

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: This becomes difficult to understand at lines 61-64 since you extract a corner from marker**s**.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        const auto corner2 = markers[2];
+        const auto corner3 = markers[3];
+
+        // Fit line between two diagonal corners
+        const auto backDiagonal{ fitLine(corner2, corner0) };
+        const auto forwardDiagonal{ fitLine(corner3, corner1) };
+
+        // Find intersection of the two lines
+        // a1*x + b1 = a2*x + b2
+        // x = (b2-b1) / (a1-a2)
```
SatjaSivcev, line 72: Why don't you have the equation for y (yCoord) as well?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        const auto xCoord =
+            (forwardDiagonal.intercept - backDiagonal.intercept) / (backDiagonal.slope - forwardDiagonal.slope);
+        const auto yCoord = backDiagonal.slope * xCoord + backDiagonal.intercept;
```
SatjaSivcev, line 75: Not a must, but perhaps a good idea. I would consider using 'u' and 'v' for image (2D) coordinates, and 'x' 'y' 'z' for 3D coordinates. This will help to avoid any confusion on what is xCoord for example.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+            return {};
+        }
+
+        std::vector<cv::Point2f> markerCenters2D;
+        markerCenters2D.reserve(markers.size());
+
+        const auto corner0 = markers[0];
+        const auto corner1 = markers[1];
+        const auto corner2 = markers[2];
+        const auto corner3 = markers[3];
```
SatjaSivcev, line 64: markerCorner0 etc.?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        // x = (b2-b1) / (a1-a2)
+        const auto xCoord =
+            (forwardDiagonal.intercept - backDiagonal.intercept) / (backDiagonal.slope - forwardDiagonal.slope);
+        const auto yCoord = backDiagonal.slope * xCoord + backDiagonal.intercept;
+        markerCenters2D.emplace_back(cv::Point2f(xCoord, yCoord));
+
+        return markerCenters2D;
+    }
+
+    std::vector<cv::Point2f> get2DMarkerCorners(const aruconano::Marker &markers)
```
SatjaSivcev, line 81: Why is 'markers' plural? Isn't it a single marker?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: Or can this actually hold multiple markers and then the corners are stored like [marker1corner1, marker1corner2, marker1corner3, marker1corner4, marker2corner1, marker2corner2,...]?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        const int width = pointCloud.width();
+        const int height = pointCloud.height();
+        const auto points = pointCloud.copyPointsXYZ();
+
+        std::vector<cv::Point3d> markerPoints3D;
+        markerPoints3D.reserve(markerPoints2D.size());
+        for(const auto &point2D : markerPoints2D)
+        {
+            const float x = point2D.x;
+            const float y = point2D.y;
```
SatjaSivcev, line 117: Not a must, but perhaps a good idea. I would consider using 'u' and 'v' for image (2D) coordinates, and 'x' 'y' 'z' for 3D coordinates.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+            const int x1 = clamp(static_cast<int>(std::floor(x)), 0, width - 1);
+            const int x2 = clamp(static_cast<int>(std::ceil(x)), 0, width - 1);
+            const int y1 = clamp(static_cast<int>(std::floor(y)), 0, height - 1);
+            const int y2 = clamp(static_cast<int>(std::ceil(y)), 0, height - 1);
+
+            // estimate the 3d center of the marker using bilinear interpolation
```
SatjaSivcev, line 124: Use -ing continuous tense, and this comment explains the whole function so it should be moved up where the function definition starts.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+    Zivid::Matrix4x4 getMarkerPose(const Zivid::PointCloud &pointCloud, const aruconano::Marker &markers)
+    {
+        // Extracting 2d corners and estimateing 2d center
+        const auto corners2D = get2DMarkerCorners(markers);
+        const auto center2D = estimate2DMarkerCenters(markers);
+
+        // estimating 3d corners and center from 2d data
+        const auto corners3D = estimate3DMarkerPoints(pointCloud, corners2D);
+        const auto center3D = estimate3DMarkerPoints(pointCloud, center2D);
```
SatjaSivcev, line 196: Write a comment somewhere that 2D coordinates of marker corners / centers are given as real numbers and therefore it is not straightforward to extract the corresponding 3D coordinates. For this reason, it is necessary to do bilinear interpolation.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+            const int x2 = clamp(static_cast<int>(std::ceil(x)), 0, width - 1);
+            const int y1 = clamp(static_cast<int>(std::floor(y)), 0, height - 1);
+            const int y2 = clamp(static_cast<int>(std::ceil(y)), 0, height - 1);
+
+            // estimate the 3d center of the marker using bilinear interpolation
+            // See https://en.wikipedia.org/wiki/Bilinear_interpolation
+            const auto q11 = points(y1, x1);
+            const auto q12 = points(y2, x1);
+            const auto q21 = points(y1, x2);
+            const auto q22 = points(y2, x2);
```
SatjaSivcev, line 129: Add a comment: // Four known points next to the point at which we are interpolating

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        markerPoints3D.reserve(markerPoints2D.size());
+        for(const auto &point2D : markerPoints2D)
+        {
+            const float x = point2D.x;
+            const float y = point2D.y;
+
+            const int x1 = clamp(static_cast<int>(std::floor(x)), 0, width - 1);
+            const int x2 = clamp(static_cast<int>(std::ceil(x)), 0, width - 1);
+            const int y1 = clamp(static_cast<int>(std::floor(y)), 0, height - 1);
+            const int y2 = clamp(static_cast<int>(std::ceil(y)), 0, height - 1);
```
SatjaSivcev, line 122: Add a comment: // Getting coordinates for the four known points next to the interpolating point

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+            return {};
+        }
+
+        const int width = pointCloud.width();
+        const int height = pointCloud.height();
+        const auto points = pointCloud.copyPointsXYZ();
+
+        std::vector<cv::Point3d> markerPoints3D;
+        markerPoints3D.reserve(markerPoints2D.size());
+        for(const auto &point2D : markerPoints2D)
```
SatjaSivcev, line 114: This is where the bilinear interpolation start. Write that in the comment and move 'see https://en.wikipedia.org/wiki/Bilinear_interpolation' here

SatjaSivcev: When this is done, then I think it makes sense to have simple variables such as x y a b that are equivalent to those at the referenced website.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+            // See https://en.wikipedia.org/wiki/Bilinear_interpolation
+            const auto q11 = points(y1, x1);
+            const auto q12 = points(y2, x1);
+            const auto q21 = points(y1, x2);
+            const auto q22 = points(y2, x2);
+
+            const float a = (x2 - x) / (x2 - x1);
+            const float b = (x - x1) / (x2 - x1);
+
+            // linear interpolation in x direction
```
SatjaSivcev, line 134: // linear interpolation in x direction vs // linear interpolation in y-axis - make it consistent, use axis or direction?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+            // estimate the 3d center of the marker using bilinear interpolation
+            // See https://en.wikipedia.org/wiki/Bilinear_interpolation
+            const auto q11 = points(y1, x1);
+            const auto q12 = points(y2, x1);
+            const auto q21 = points(y1, x2);
+            const auto q22 = points(y2, x2);
+
+            const float a = (x2 - x) / (x2 - x1);
+            const float b = (x - x1) / (x2 - x1);
```
SatjaSivcev, line 132: I think you should get rid of a and b and put the values directly in the equations. This is how the function is defined on wiki.

Same for c and d

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev, line 50: Why not esto,ate2DMarkerCenter() - singular?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        auto poseCenter = center2D[0];
+        auto xAxisPoint = cv::Point2f(((corners2D[3].x + corners2D[0].x) / 2), ((corners2D[3].y + corners2D[0].y) / 2));
+        auto yAxisPoint = cv::Point2f(((corners2D[2].x + corners2D[3].x) / 2), ((corners2D[2].y + corners2D[3].y) / 2));
+
+        cv::circle(bgr, poseCenter, 2, cv::Scalar(255, 0, 0), 2);
+        cv::arrowedLine(bgr, poseCenter, xAxisPoint, cv::Scalar(0, 0, 255), 2);
+        cv::arrowedLine(bgr, poseCenter, yAxisPoint, cv::Scalar(0, 255, 0), 2);
+
+        cv::imshow("Pose of aruco marker", bgr);
```
SatjaSivcev, line 184: 'ArUco' - use this style everywhere

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        cv::waitKey(0);
+    }
+
+    Zivid::Matrix4x4 getMarkerPose(const Zivid::PointCloud &pointCloud, const aruconano::Marker &markers)
+    {
+        // Extracting 2d corners and estimateing 2d center
+        const auto corners2D = get2DMarkerCorners(markers);
+        const auto center2D = estimate2DMarkerCenters(markers);
+
+        // estimating 3d corners and center from 2d data
```
SatjaSivcev, line 194: Capital E

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        cv::arrowedLine(bgr, poseCenter, xAxisPoint, cv::Scalar(0, 0, 255), 2);
+        cv::arrowedLine(bgr, poseCenter, yAxisPoint, cv::Scalar(0, 255, 0), 2);
+
+        cv::imshow("Pose of aruco marker", bgr);
+        cv::waitKey(0);
+    }
+
+    Zivid::Matrix4x4 getMarkerPose(const Zivid::PointCloud &pointCloud, const aruconano::Marker &markers)
+    {
+        // Extracting 2d corners and estimateing 2d center
```
SatjaSivcev, line 190: Do we use 2d pr 2D elsewhere?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        const auto corners2D = get2DMarkerCorners(markers);
+        const auto center2D = estimate2DMarkerCenters(markers);
+
+        // estimating 3d corners and center from 2d data
+        const auto corners3D = estimate3DMarkerPoints(pointCloud, corners2D);
+        const auto center3D = estimate3DMarkerPoints(pointCloud, center2D);
+
+        // extracting origin and calcualting normal vectors for x-, y- and z-axis
```
SatjaSivcev, line 198: Capital E

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        cv::circle(bgr, poseCenter, 2, cv::Scalar(255, 0, 0), 2);
+        cv::arrowedLine(bgr, poseCenter, xAxisPoint, cv::Scalar(0, 0, 255), 2);
+        cv::arrowedLine(bgr, poseCenter, yAxisPoint, cv::Scalar(0, 255, 0), 2);
+
+        cv::imshow("Pose of aruco marker", bgr);
+        cv::waitKey(0);
+    }
+
+    Zivid::Matrix4x4 getMarkerPose(const Zivid::PointCloud &pointCloud, const aruconano::Marker &markers)
```
SatjaSivcev, line 188: rename to getAndDisplayMarkerPose if you display in this function. Alternative is to take displayMarkerPose() as a separate function and call it outside this one.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        auto rotationMatrix = cv::Mat(3, 3, CV_64F);
+        for(int i = 0; i < 3; ++i)
+        {
+            rotationMatrix.at<double>(i, 0) = u[i];
+            rotationMatrix.at<double>(i, 1) = v[i];
+            rotationMatrix.at<double>(i, 2) = unitNormal[i];
+        }
+
+        auto transform = transformationMatrix(rotationMatrix, origin);
```
SatjaSivcev, line 222: this can go directly to return - line 227

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+    {
+        Zivid::Application zivid;
+
+        const auto dataFile = std::string(ZIVID_SAMPLE_DATA_DIR) + "/ArucoSample.zdf";
+        std::cout << "Reading ZDF frame from file: " << dataFile << std::endl;
+        const Zivid::Frame frame = Zivid::Frame(dataFile);
+        auto pointCloud = frame.pointCloud();
+
+        std::cout << "Converting to OpenCV BGR image format" << std::endl;
+        cv::Mat image = pointCloudToCvBGR(pointCloud);
```
SatjaSivcev, line 283: cv::Mat **bgr** = pointCloudToCvBGR(pointCloud);

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        std::cout << "Detecting aruco markers in 2D image" << std::endl;
```
SatjaSivcev, line 285: 'ArUco'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        auto markers = aruconano::MarkerDetector::detect(image);
```
SatjaSivcev, line 286: singular - marker?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        std::cout << "Drawing marker corners" << std::endl;
+        for(const auto &marker : markers)
+        {
+            marker.draw(image);
+        }
+        std::cout << "Displaying detected markers" << std::endl;
+        cv::imshow("Detected aruco markers", image);
+        cv::waitKey(0);
```
SatjaSivcev, line 295: make a function? displayMarkers()

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        std::cout << "Detecting aruco marker poses" << std::endl;
```
SatjaSivcev, line 297: 'ArUco'

robertkt1:
What is the benefit of using a new aruco lib compared to cv aruco?

Should we also include something about aruco library as we do with eigen, pcl and cv in the readme? How to build with or without aruco depending. For eigen we did: -DUSE_EIGEN3=OFF

What happens if you try to use cmake in windows without pointing to aruco lib. Will it give the same nice "please point to aruco lib" as for eigen, cv and pcl?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev, line 188: use marker (singluar) as the argument is markers[0] in main - which is good

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+int main()
+{
+    try
+    {
+        Zivid::Application zivid;
+
+        const auto dataFile = std::string(ZIVID_SAMPLE_DATA_DIR) + "/ArucoSample.zdf";
+        std::cout << "Reading ZDF frame from file: " << dataFile << std::endl;
+        const Zivid::Frame frame = Zivid::Frame(dataFile);
+        auto pointCloud = frame.pointCloud();
```
SatjaSivcev, line 280: const auto pointCloud = Zivid::Frame(dataFile).pointCloud();

You don't use frame and I think you don't modify pointCloud so it can be const?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        }
+        std::cout << "Displaying detected markers" << std::endl;
+        cv::imshow("Detected aruco markers", image);
+        cv::waitKey(0);
+
+        std::cout << "Detecting aruco marker poses" << std::endl;
+        const auto transform = getMarkerPose(pointCloud, markers[0]);
+
+        std::cout << "Aruco marker transformation matrix:" << std::endl;
+        std::cout << transform << std::endl;
```
SatjaSivcev, line 301: Delete empty line

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+            marker.draw(image);
+        }
+        std::cout << "Displaying detected markers" << std::endl;
+        cv::imshow("Detected aruco markers", image);
+        cv::waitKey(0);
+
+        std::cout << "Detecting aruco marker poses" << std::endl;
+        const auto transform = getMarkerPose(pointCloud, markers[0]);
+
+        std::cout << "Aruco marker transformation matrix:" << std::endl;
```
SatjaSivcev, line 300: ArUco

SatjaSivcev, line 300: Change to 'ArUco marker pose in camera frame'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+        for(const auto &marker : markers)
+        {
+            marker.draw(image);
+        }
+        std::cout << "Displaying detected markers" << std::endl;
+        cv::imshow("Detected aruco markers", image);
+        cv::waitKey(0);
+
+        std::cout << "Detecting aruco marker poses" << std::endl;
+        const auto transform = getMarkerPose(pointCloud, markers[0]);
```
SatjaSivcev, line 298: Now that I have a better overview, we should rename transform and invertedTransform to transformMarkerToCamera and transformCameraToMarker

transformAToB is Pose of A in frame of B

See https://github.com/zivid/zivid-cpp-samples/blob/7be0801ade2e704860a67f3d3985fe33e6b754da/source/Applications/Advanced/HandEyeCalibration/UtilizeEyeInHandCalibration/UtilizeEyeInHandCalibration.cpp#L150 for reference

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+    }
+} // namespace
+
+int main()
+{
+    try
+    {
+        Zivid::Application zivid;
+
+        const auto dataFile = std::string(ZIVID_SAMPLE_DATA_DIR) + "/ArucoSample.zdf";
```
SatjaSivcev, line 277: I think we can improve names for "ArucoSample.zdf" and "transformedArucoSample.zdf"
- 'Sample': this is not needed imo - what value does it give?
- We should indicate somehow the origin of the point cloud: e.g. ArUcoInCameraOrigin and ArUcoInArUcoOrigin? (the latter maybe doesn't sound well...), or PointCloudInCameraOrigin and PointCloudInMarkerOrigin?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        std::cout << "Detecting aruco markers in 2D image" << std::endl;
+        auto markers = aruconano::MarkerDetector::detect(image);
+
+        std::cout << "Drawing marker corners" << std::endl;
+        for(const auto &marker : markers)
+        {
+            marker.draw(image);
+        }
+        std::cout << "Displaying detected markers" << std::endl;
```
SatjaSivcev, line 293: I think this printout should be placed instead of "Drawing marker corners" which is not needed (the first one is sufficient)

SatjaSivcev (changes requested):
Changes requested

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: You mean "Getting 3d coordinates for the four points whose 2D coordinates are known"? x1, y1, etc are the pixel coordinates and hence we get the 3D coordinates with points(x1,y1). Added a sentence explaining this

README.md
chrisasc: I agree that 'Transform' can mean anything, but with the rest of the name (pointCloudUsingArucoMarker), i think it puts it in context. 

If we are going to change it then I like 'TransformPointCloudViaArucoMarker' or 'TransformPointCloudUsingArucoMarker'.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Ci takes care of this. The space is to separate between different types of libraries we include (local header files, standard library, or 3rd party libraries). We do have the same spacing in all our samples.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: > @chrisasc Have you introduced any // NOLINT that is not already used in the samples? If not sure, check please.

No, I removed those I added earlier. These are from the multi-camera samples that Torbjørn made

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Actually the code can detect multiple markers and for each marker find the corner and pose. But, from our discussion of this and the ROI sample, i can edit it to only detect a single marker.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: One diagonal goes from left to right (if you picture the 4 corners in front of you), while the other diagonal goes from right to left. I am not able to come up with better names for them:/

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: putted inside the function. This is what we do in other c++ samples.

README.md
robertkt1: I agree we should mention pointcloud and aruco in the name. I like both suggestions from christians last comment.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Because its the equation of a line: y = a1*x +b1. But added it now

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Because it can take in multiple aruco markers. But i can modify the code to only accept one.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: I think the real reason we want to do bilinear interpolation is to get better results. We could always just convert the center/corner floats to int and say that this is good enough. But then the resulting transformation would be worse.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: agree, done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: agree, done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: because the function can take in multiple aruco markers. But i can change that (see earlier comments about this)

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: we use 2D. Also changed from 3d to 3D

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: better to call it getAndDisplayMarkerPose, since it requires 2D corner and centers to display the pose. These are not returned from the function

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: yes it can. Moved it

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: looks better, done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: i agree, its not necessary to have all those comments. done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: I cannot be const, since me actually modify it when we use pointCloud.transform().

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: I agree, that way there is no doubt what kind of transformation you are doing. Nice one!

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: The name of the file is not finsihed. I just needed to call it something. So good that you start a discussion on the naming of the file!

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: I removed the other print statement when i changed that diisplayAruCOMarker to a function. Now its only one print statement (which corresponds to the actual function name as well)

README.md
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: ah, changed it to bgr

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

README.md
chrisasc: agree, done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

chrisasc:
Went through your comments Satja. Should this sample only take a single marker or multiple? For simplicity it should be only a single marker, but it would be cool to detect multiple. Depending on the outcome i will modify the name of the functions to either singular or plural.

SatjaSivcev:
> Went through your comments Satja. Should this sample only take a single marker or multiple? For simplicity it should be only a single marker, but it would be cool to detect multiple. Depending on the outcome i will modify the name of the functions to either singular or plural.

I think we should chose simplicity for now.

README.md
SatjaSivcev: It is better now. nitpick - we don't use 'a' before 'point cloud' in other sample descriptions, we agreed adding all articles would be verbose.

README.md
SatjaSivcev: I think that is sufficient enough. Whatever you decide don't implement before the review is finished.

SatjaSivcev:
> @chrisasc Have you introduced any // NOLINT that is not already used in the samples? If not sure, check please.

@chrisasc bump

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: diagonal02 and diagonal 13? since you use corners 0 1 2 3? Or diagonalCorner0ToCorner2 - these are some ideas to start with. I think 'back' and 'forward' doesn't reveal anything.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: What do you suggest then? To write something in that line, e.g. '2D coordinates of marker corners / centers are given as real numbers and therefore it is not straightforward to extract the corresponding 3D coordinates. Bilinear interpolation is used to improve the estimation of the 3D coordinates'?

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: you didn't write 'next'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: you didn't write 'next'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: You wrote '// Getting 3D coordinates for the four points whose 2D coordinates are known' which is a comment that depends on a previous comment (not standalone). I think '// Getting 3D coordinates for the four points next to the interpolation point' is better

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,321 @@
...
+#include <opencv2/highgui/highgui.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+
+#include <aruco_nano.h>
+#include <cmath>
+#include <iostream>
+
+namespace
+{
+    struct Line2d
```
SatjaSivcev, line 20: line2D

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,321 @@
...
+        double slope;
+        double intercept;
+    };
+
+    int clamp(const int n, const int lower, const int upper)
+    {
+        return std::max(lower, std::min(n, upper));
+    }
+
+    Line2d fitLine(const cv::Point2d &point1, const cv::Point2d &point2)
```
SatjaSivcev, line 31: 2D

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,321 @@
...
+        cv::Mat bgr = pointCloudToCvBGR(pointCloud);
+
+        std::cout << "Detecting ArUco marker in 2D image" << std::endl;
+        auto markers = aruconano::MarkerDetector::detect(bgr);
+
+        std::cout << "Displaying detected ArUco marker" << std::endl;
+        displayArUcoMarker(bgr, markers);
+
+        std::cout << "Estimating detected ArUco marker's pose" << std::endl;
+        const auto transformMarkerToCamera = getAndDisplayMarkerPose(pointCloud, markers[0]);
```
SatjaSivcev, line 304: In line 303 add something in the comment with regards to 'displaying'.
Maybe clarify in line 300 that it is displayed in 2D image, and here that it is displayed in the point cloud?

SatjaSivcev, line 304: getPoseAndDisplayMarker is a better name - you are displaying Marker not Marker Pose.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
```diff
@@ -0,0 +1,321 @@
...
+        auto pointCloud = frame.pointCloud();
+
+        std::cout << "Converting to OpenCV BGR image format" << std::endl;
+        cv::Mat bgr = pointCloudToCvBGR(pointCloud);
+
+        std::cout << "Detecting ArUco marker in 2D image" << std::endl;
+        auto markers = aruconano::MarkerDetector::detect(bgr);
+
+        std::cout << "Displaying detected ArUco marker" << std::endl;
+        displayArUcoMarker(bgr, markers);
```
SatjaSivcev, line 301: There is a function called displayMarkerPose which doesn't have 'ArUcO' in it. I prefer having 'ArUco' in general but since  e.g. aruconano::MarkerDetector::detect(bgr) doesn't have 'ArUco' in the name, maybe we should abandon it as well. In any case we should have consistency between displayMarker and displayMarkerPose.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: Then use ' auto pointCloud = Zivid::Frame(dataFile).pointCloud();' the point was to save one line (not have frame as a variable at all since you don't use it)

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: @chrisasc bump

chrisasc:
> > @chrisasc Have you introduced any // NOLINT that is not already used in the samples? If not sure, check please.
> 
> @chrisasc bump

No, I have not. The one that are introduced are from Torbjørn

README.md
chrisasc: OK, i wont do change it as of now. I created a task for remaining me in the PR message on the top of the page.

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: then your is better, changed to that

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: done

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Personally, I think the code looks better with a and b, but to be consistent with the equation I can change it. See image below for visual representation of the change
image

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Both of the displays are in 2D as it is quite difficult to display it in 3D

Display of the detected marker: 
image

Display of ArUco pose:
image

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: I also prefer to have ArUco included, so changed the function name to 'displayArUcoMarkerPose()' and 'displayDetectedArUcoMarker()'

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
chrisasc: Chaged the print statement to 
'std::cout << "Displaying detected ArUco marker in 2D image" << std::endl;' and
'std::cout << "Estimating and displaying detected ArUco marker's pose" << std::endl;' and

README.md
SatjaSivcev: I am implementing this in the other PR: https://github.com/zivid/zivid-cpp-samples/pull/165

source/Applications/Advanced/TransformViaArucoMarker/TransformViaArucoMarker.cpp
SatjaSivcev: I am implementing this in the other PR: #165

PR #157, opened 2021-02-08
Modify ReadPCLVis3D sample to use .pcd from SampleData

chrisasc:
Updated sample to use .pcd from SampleData instead of your own point cloud. This makes the sample independent of CaptureWritePCLVis3D.

### DIFF
--- a/source/Applications/Basic/Visualization/ReadPCLVis3D/ReadPCLVis3D.cpp
+++ b/source/Applications/Basic/Visualization/ReadPCLVis3D/ReadPCLVis3D.cpp
@@ -1,6 +1,6 @@
 /*
 This example shows how to read point cloud from PCL file and visualize it.
-To get a Zivid point cloud in PCD file format, run CaptureWritePCLVis3D sample.
+The PCD file for this sample can be found under the main instructions for Zivid samples.
 */
 
 #include <Zivid/Zivid.h>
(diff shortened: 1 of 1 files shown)


robertkt1:
Not sure if the sample description at the top of file makes sense anymore. "To get a Zivid point cloud in PCD file format, run CaptureWritePCLVis3D sample."

If you run that sample you will not get a pcd file that you can use in this sample. In readZDF we advise them to our main page to get instructions for SampleData. Maybe we should do something similar here?

SatjaSivcev:
> Not sure if the sample description at the top of file makes sense anymore. "To get a Zivid point cloud in PCD file format, run CaptureWritePCLVis3D sample."
> 
> If you run that sample you will not get a pcd file that you can use in this sample. In readZDF we advise them to our main page to get instructions for SampleData. Maybe we should do something similar here?

How about: The PCD file for this sample can be found under the main instructions for Zivid samples. Alternatively, run CaptureWritePCLVis3D sample.

Why can't the output of CaptureWritePCLVis3D be used @robertkt1?

robertkt1:
And also this cannot be merged until we update SampleData with a pcd file

robertkt1:
> 
> 
> > Not sure if the sample description at the top of file makes sense anymore. "To get a Zivid point cloud in PCD file format, run CaptureWritePCLVis3D sample."
> > If you run that sample you will not get a pcd file that you can use in this sample. In readZDF we advise them to our main page to get instructions for SampleData. Maybe we should do something similar here?
> 
> How about: The PCD file for this sample can be found under the main instructions for Zivid samples. Alternatively, run CaptureWritePCLVis3D sample.
> 
> Why can't the output of CaptureWritePCLVis3D be used @robertkt1?

Because CaptureWrite writes to the build directory and not to SampleData

chrisasc:
> And also this cannot be merged until we update SampleData with a pcd file

This is already updated

chrisasc:
> Not sure if the sample description at the top of file makes sense anymore. "To get a Zivid point cloud in PCD file format, run CaptureWritePCLVis3D sample."
> 
> If you run that sample you will not get a pcd file that you can use in this sample. In readZDF we advise them to our main page to get instructions for SampleData. Maybe we should do something similar here?

I agree! I will change that to something similar as ReadIterateZdf sample.

chrisasc:
Updated the comments on top of the sample and tested it on Windows and Ubuntu with the latest SampleData.
Ready for next round.

source/Applications/Basic/Visualization/ReadPCLVis3D/ReadPCLVis3D.cpp
```diff
@@ -1,6 +1,6 @@
 /*
 This example shows how to read point cloud from PCL file and visualize it.
-To get a Zivid point cloud in PCD file format, run CaptureWritePCLVis3D sample.
+The ZDF file for this sample can be found under the main instructions for Zivid samples.
```
robertkt1, line 3: Maybe a bit too similar as ReadIterateZdf? 
We don't use a ZDF in this sample

robertkt1: The rest looks good

PR #158, opened 2021-04-19
Update samples to SDK version 2.3.0

torbsorb:


README.md
```diff
@@ -82,7 +82,7 @@ There are two main categories of samples: **Camera** and **Applications**. The s
 ## Instructions
 
 1. [**Install Zivid Software**](https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/59080712/Zivid+Software+Installation).
-Note: The samples require Zivid SDK v2 (minor version 2.2 or newer).
+Note: The samples require Zivid SDK v2 (minor version 2.3 or newer).
```
SatjaSivcev, line 85: This change is not needed.

source/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ project(ZividCppSamples)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 include(CompilerOptions)
 
-set(ZIVID_VERSION 2.2.0)
+set(ZIVID_VERSION 2.3.0)
```
SatjaSivcev, line 8: Is this in conflict with our statement: Note: The samples require Zivid SDK v2 (minor version 2.2 or newer)?
I guess, we need to add stg like. To use version x.y.z change CMakeLists to x.y.z?

README.md
torbsorb: True in theory, but it will fail on configuration because CMakeLists states 2.3.0

SatjaSivcev (changes requested):
minor changes

README.md
SatjaSivcev: Add a sentence to clarify then, see https://github.com/zivid/zivid-cpp-samples/pull/158#discussion_r615675488

torbsorb: I'm wrong. This seems to work as a "minimum" version match

source/CMakeLists.txt
torbsorb, line 8: Based on the above we should not update this one

torbsorb: I checked and this check in CMake is not EXACT. Not entirely sure what it means (although it's explained here: https://cmake.org/cmake/help/latest/command/find_package.html#version-selection). In any case. I tested, and it works with both 2.2 and 2.3 if we leave it at 2.2. If we update to 2.3 then it only works with 2.3. It looks like it checks for minimum compatibility.

SatjaSivcev: That is great.

SatjaSivcev (approved):
LGTM

PR #160, opened 2021-07-19
Add Warmup Sample

SatjaSivcev:


ajlandau:
Changed it to no longer use namespace
did not make other changes

Do we want to print temperature anywhere? 

@robertkt1

SatjaSivcev:
> Changed it to no longer use namespace
> did not make other changes
> 
> Do we want to print temperature anywhere? 
> 
> @robertkt1 

Why not use namespace? Consult SE about it? I think they recommend to use it in general.

I vote for not printing temperature. Simplicity.

eskaur:
> > Changed it to no longer use namespace
> > did not make other changes
> > Do we want to print temperature anywhere?
> > @robertkt1
> 
> Why not use namespace? Consult SE about it? I think they recommend to use it in general.
> 
> I vote for not printing temperature. Simplicity.

If you define a function in a cpp file that is only used in that file, then that function should definitely be in an (anonymous) namespace like it was in your commit. What @ajlandau 's commit did was not really about removing the namespace, it was removing the function and inlining its content (which had the side-effect of making the namespace empty, so it could be removed).

In this case I think the function is unnecessary, and agree with Amy's change.

source/Camera/InfoUtilOther/Warmup/Warmup.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    try
+    {
+        Zivid::Application zivid;
+
+        std::cout << "Connecting to camera" << std::endl;
+        auto camera = zivid.connectCamera();
+
+        auto warmupTime = std::chrono::minutes(10);
+        auto captureCycle = std::chrono::seconds(5);
```
robertkt1, line 24: can also be const?

source/Camera/InfoUtilOther/Warmup/Warmup.cpp
```diff
@@ -0,0 +1,70 @@
...
+            auto captureTime = afterCapture - beforeCapture;
+            if(captureTime < captureCycle)
+            {
+                std::this_thread::sleep_for(captureCycle - captureTime);
+            }
+            else
+            {
+                std::cout << "Your capture time is longer than your desired capture cycle."
+                          << "Please increase the desired capture cycle." << std::endl;
+            }
```
robertkt1, line 54: Not sure if I like this. We say "please increase the desired capture cycle", but the loops keeps on going. If they don't change the warm up time this is going to print every 5 seconds for 10 minutes. 
Not sure if that is a problem, but just thinking out loud here.

robertkt1: If we are using capture assistant we could have at a variable for max capture time and compared that to cycle time. 
But since cycle time is 5s by default and capture assistant i 1s by default it should work out-of-the-box and if they do the modification they will be aware of what is happening. 
Don't think my first comment is important..

source/Camera/InfoUtilOther/Warmup/Warmup.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+        const auto beforeWarmup = HighResClock::now();
+
+        while(HighResClock::now() - beforeWarmup < warmupTime)
+        {
+            const auto beforeCapture = HighResClock::now();
+
+            camera.capture(settings);
+
+            const auto afterCapture = HighResClock::now();
```
robertkt1, line 44: Not sure if beforeCapture and AfterCapture should be const? They change for each iteration. (Not sure what the standard is for const in loops)

robertkt1: If they are const, then all variables in the loop could be const?

ajlandau: Lets maybe ask SE? I do not know.

ajlandau:
@robertkt1 Fixed it up. 

pls check it out

robertkt1 (commented):
Add sample to readme. 

Rest LGTM

source/Camera/InfoUtilOther/Warmup/Warmup.cpp
```diff
@@ -0,0 +1,72 @@
...
```
robertkt1, line 44: can remove empty lines?

source/Camera/InfoUtilOther/Warmup/Warmup.cpp
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            const auto remainingTime = warmupTime - (HighResClock::now() - beforeWarmup);
+
+            const auto remainingTimeMinutes = std::chrono::duration_cast<std::chrono::minutes>(remainingTime);
+            const auto remainingTimeSeconds =
+                std::chrono::duration_cast<std::chrono::seconds>(remainingTime - remainingTimeMinutes);
+            std::cout << "Remaining time: " << remainingTimeMinutes.count() << " minutes, "
+                      << remainingTimeSeconds.count() << " seconds" << std::endl;
+        }
```
robertkt1, line 63: Maybe add print: "Warm up completed"?

robertkt1 (approved):
Assume you have tested on both windows and lnux?
LGTM

PR #161, opened 2021-07-30
Add multi camera capture sample

robertkt1:


source/Applications/Advanced/MultiCamera/MultiCameraCaptureInParallel/MultiCameraCaptureInParallel.cpp
```diff
@@ -0,0 +1,76 @@
...
+        };
+        const auto settings = Zivid::CaptureAssistant::suggestSettings(camera, parameters);
+
+        // Capturing frame
+        auto frame = camera.capture(settings);
+
+        const auto dataFile = "Frame_" + camera.info().serialNumber().value() + ".zdf";
+
+        // Saving frame to file
+        std::unique_lock<std::mutex> lock(m);
```
torbsorb, line 29: Where is the unlock?

torbsorb: Ahh.. this is for single execution, and you're just protecting file write? Question 1: Is it required to protect file write? Question 2: Should we still explicitly unlock the mutex?

source/Applications/Advanced/MultiCamera/README.md
```diff
@@ -1,5 +1,10 @@
 # Multi Camera
 
+To use multiple Zivid Two cameras, set custom IP configuration. See guide for [multiple zivid cameras](https://support.zivid.com/latest/academy/camera/multiple-zivid-cameras.html?highlight=multi%20camera) in our Knowledge Base for complete guide.
```
torbsorb, line 3: "?highlight=multi%20camera" do we really want to highlight this text in the browser?

torbsorb (approved):
It's probably ok as it is, although I had a couple of questions.

source/Applications/Advanced/MultiCamera/MultiCameraCaptureInParallel/MultiCameraCaptureInParallel.cpp
torbsorb: I see that the destructor of std:unique_lock unlocks the mutex, so that's quite explicit.

robertkt1: 1) Yeah writing to file failed every time I didn't do mutex. 
2) yes it unlocks as you say

source/Applications/Advanced/MultiCamera/README.md
robertkt1: hehe we probably dont want that

robertkt1: Changed the link to ip config page instead.

source/Camera/Advanced/MultiCameraCaptureInParallel/MultiCameraCaptureInParallel.cpp
```diff
@@ -0,0 +1,74 @@
...
+        const auto settings = Zivid::CaptureAssistant::suggestSettings(camera, parameters);
+
+        // Capturing frame
+        const auto frame = camera.capture(settings);
+
+        const auto dataFile = "Frame_" + camera.info().serialNumber().value() + ".zdf";
+
+        // Saving frame to file
+        std::unique_lock<std::mutex> lock(m);
+        frame.save("Frame_" + camera.info().serialNumber().value() + ".zdf");
```
apartridge, line 30: Is there a reason that you add a mutex around frame.save()? I don't think that should be needed.
You have made the variable `dataFile` above, but you are not using it here.

source/Camera/Advanced/MultiCameraCaptureSequentially/MultiCameraCaptureSequentially.cpp
```diff
@@ -0,0 +1,49 @@
...
+                Zivid::CaptureAssistant::SuggestSettingsParameters::AmbientLightFrequency::none,
+                Zivid::CaptureAssistant::SuggestSettingsParameters::MaxCaptureTime{ std::chrono::milliseconds{ 800 } }
+            };
+            const auto settings = Zivid::CaptureAssistant::suggestSettings(camera, parameters);
+            auto frame = camera.capture(settings);
+
+            const auto dataFile = "Frame_" + camera.info().serialNumber().value() + ".zdf";
+
+            std::cout << "Saving frame to file: " << dataFile << std::endl;
+            frame.save("Frame_" + camera.info().serialNumber().value() + ".zdf");
```
apartridge, line 39: You can use dataFile here

source/Camera/Advanced/MultiCameraCaptureInParallel/MultiCameraCaptureInParallel.cpp
```diff
@@ -0,0 +1,74 @@
...
+        std::mutex fileSaveMutex;
+
+        for(auto &camera : cameras)
+        {
+            std::cout << "Starting task with camera: " << camera.info().serialNumber().value() << std::endl;
+            container.emplace_back(
+                std::async(std::launch::async, captureInThread, std::ref(camera), std::ref(fileSaveMutex)));
+        }
+
+        for(unsigned int i = 0; i < cameras.size(); ++i)
```
apartridge, line 61: size_t i

source/Camera/Advanced/MultiCameraCaptureInParallel/MultiCameraCaptureInParallel.cpp
knzivid: It crashes without a mutex. netCDF doc says not all netCDF calls are thread safe.

knzivid: I don't know exactly what crashes. I only see a netCDF function on the top of the call stack when this segfault happens

robertkt1: on my laptop it crashes without the mutex. Not sure the reason

apartridge: @robertkt1 OK; can you make an SDK issue for this?

robertkt1: @apartridge I could, but I am not sure what the issue would be. If Karthik says not all netCDF calls are thread safe then I don't see this as a bug? 
Or is it that we want all our (one-camera-)api to be thread safe?
I haven't tested it, but if the one-camera-api should be thread safe I assume we want to test downsampling and transformation in multi threads as well.

apartridge: Our official position is that one can use multiple cameras at the same time from different threads, as long as it is one thread per camera (and that you connect to cameras sequentially). We don't say anything in particular about saving frame from multiple cameras at the same time, but from an user perspective I would expect that to work (same as other frame/point cloud operations like copy data to CPU, downsampling, transform etc). So I think we should make a bug for this, so that we can improve multi-threading behavior in the future. If netCDF itself is not thread safe, that just means we must add locking on SDK side. NetCDF should not dictate our multi-threading policy.

PR #163, opened 2021-08-16
Add ToneMapping to CaptureHDRCompleteSettings

SatjaSivcev:


chrisasc (approved):

apartridge:
We are missing at least Gamma and Acquisition::Brightness in this sample (if the idea is to include every setting).

SatjaSivcev:
> We are missing at least Gamma and Acquisition::Brightness in this sample (if the idea is to include every setting).

Brightness we have (line 67):

const auto baseAcquisition = Zivid::Settings::Acquisition{ Zivid::Settings::Acquisition::Brightness{ 1.8 } };

PR #166, opened 2021-09-23
Add TransformPointCloudViaArucoMarker sample

SatjaSivcev:
The sample reads a point cloud, detects an aruco marker, and estimates
its pose. Then, it transforms the point cloud to the marker coordinate
frame and saves the transformed point cloud.

Continuation of https://github.com/zivid/zivid-cpp-samples/pull/153

### DIFF
--- a/source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
+++ b/source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
@@ -1,0 +1,311 @@
+/*
+This example shows how to transform a point cloud from camera to ArUco Marker coordinate frame
+by estimating the marker's pose from the point cloud.
+*/
+
+#include <Zivid/Experimental/Calibration.h>
+#include <Zivid/Zivid.h>
+
+#include <algorithm>
+#include <opencv2/core/core.hpp>
+#include <opencv2/highgui/highgui.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+
+#include <opencv2/aruco.hpp>
+
+#include <cmath>
+#include <iostream>
+
+namespace
+{
+    struct Line2d
+    {
+        double slope;
+        double intercept;
+    };
+
+    float clamp(const float n, const float lower, const float upper)
+    {
+        return std::max(lower, std::min(n, upper));
+    }
+
+    Line2d fitLine(const cv::Point2d &point1, const cv::Point2d &point2)
+    {
+        // Fitting a line y=a*x + b to 2 points
+        const auto slope{ (point2.y - point1.y) / (point2.x - point1.x) };
+        const auto intercept{ point1.y - slope * point1.x };
+        return { slope, intercept };
+    };
+
+    cv::Point2f estimate2DMarkerCenter(const std::vector<cv::Point2f> &markerCorners)
+    {
+        const auto markerCorner0 = markerCorners[0];
+        const auto markerCorner1 = markerCorners[1];
+        const auto markerCorner2 = markerCorners[2];
+        const auto markerCorner3 = markerCorners[3];
+
+        // Fitting line between two diagonal marker corners
+        const auto backDiagonal{ fitLine(markerCorner2, markerCorner0) };
+        const auto forwardDiagonal{ fitLine(markerCorner3, markerCorner1) };
+
+        // Finding intersection of the two lines
+        // a1*x + b1 = a2*x + b2
+        // x = (b2-b1) / (a1-a2)
+        // y = a1*x + b1
+        const auto xCoordinate =
+            (forwardDiagonal.intercept - backDiagonal.intercept) / (backDiagonal.slope - forwardDiagonal.slope);
+        const auto yCoordinate = backDiagonal.slope * xCoordinate + backDiagonal.intercept;
+        auto markerCenter = cv::Point2f(xCoordinate, yCoordinate);
+
+        return markerCenter;
+    }
+
+    std::vector<cv::Point3d> estimate3DMarkerPoints(const Zivid::PointCloud &pointCloud,
+                                                    const std::vector<cv::Point2f> &markerPoints2D)
+    {
+        if(markerPoints2D.empty())
+        {
+            return {};
+        }
+
+        const float width = pointCloud.width();
+        const float height = pointCloud.height();
+        const auto points = pointCloud.copyPointsXYZ();
+
+        std::vector<cv::Point3d> markerPoints3D;
+        markerPoints3D.reserve(markerPoints2D.size());
+
+        for(const auto &point2D : markerPoints2D)
+        {
+            // Estimating the 3D center/corners of the marker using bilinear interpolation
+            // See https://en.wikipedia.org/wiki/Bilinear_interpolation
+            const auto x = point2D.x;
+            const auto y = point2D.y;
+
+            // Getting pixel coordinates for the four known pixels next to the interpolation point
+            const auto xRoundedDown = clamp(std::floor(x), 0, width - 1);
+            const auto xRoundedUp = clamp(std::ceil(x), 0, width - 1);
+            const auto yRoundedDown = clamp(std::floor(y), 0, height - 1);
+            const auto yRoundedUp = clamp(std::ceil(y), 0, height - 1);
+
+            // Getting 3D coordinates for the four points next to the interpolation point
+            const auto q11 = points(yRoundedDown, xRoundedDown);
+            const auto q12 = points(yRoundedUp, xRoundedDown);
+            const auto q21 = points(yRoundedDown, xRoundedUp);
+            const auto q22 = points(yRoundedUp, xRoundedUp);
+
+            // Linear interpolation in x direction
+            // f(x,y1) = (x2 - x)/(x2 - x1) * f(q11) + (x - x1)/(x2 - x1) * f(q21)
+            const cv::Point3f fxy1{ (xRoundedUp - x) / (xRoundedUp - xRoundedDown) * q11.x
+                                        + (x - xRoundedDown) / (xRoundedUp - xRoundedDown) * q21.x,
+                                    (xRoundedUp - x) / (xRoundedUp - xRoundedDown) * q11.y
+                                        + (x - xRoundedDown) / (xRoundedUp - xRoundedDown) * q21.y,
+                                    (xRoundedUp - x) / (xRoundedUp - xRoundedDown) * q11.z
+                                        + (x - xRoundedDown) / (xRoundedUp - xRoundedDown) * q21.z };
+            // f(x,y2) = (x2 - x)/(x2 - x1) * f(q12) + (x - x1)/(x2 - x1) * f(q22)
+            const cv::Point3f fxy2{ (xRoundedUp - x) / (xRoundedUp - xRoundedDown) * q12.x
+                                        + (x - xRoundedDown) / (xRoundedUp - xRoundedDown) * q22.x,
+                                    (xRoundedUp - x) / (xRoundedUp - xRoundedDown) * q12.y
+                                        + (x - xRoundedDown) / (xRoundedUp - xRoundedDown) * q22.y,
+                                    (xRoundedUp - x) / (xRoundedUp - xRoundedDown) * q12.z
+                                        + (x - xRoundedDown) / (xRoundedUp - xRoundedDown) * q22.z };
+
+            // Linear interpolation in y direction
+            // f(x,y) = (y2 - y)/(y2 - y1) * f(x,y1) + (y - y1)/(y2 - y1) * f(x,y2)
+            markerPoints3D.emplace_back(cv::Point3f{ (yRoundedUp - y) / (yRoundedUp - yRoundedDown) * fxy1.x
+                                                         + (y - yRoundedDown) / (yRoundedUp - yRoundedDown) * fxy2.x,
+                                                     (yRoundedUp - y) / (yRoundedUp - yRoundedDown) * fxy1.y
+                                                         + (y - yRoundedDown) / (yRoundedUp - yRoundedDown) * fxy2.y,
+                                                     (yRoundedUp - y) / (yRoundedUp - yRoundedDown) * fxy1.z
+                                                         + (y - yRoundedDown) / (yRoundedUp - yRoundedDown) * fxy2.z });
+        }
+
+        return markerPoints3D;
+    }
+
+    Zivid::Matrix4x4 transformationMatrix(const cv::Matx33d &rotationMatrix, const cv::Vec3d &translationVector)
+    {
+        auto transformMatrix = Zivid::Matrix4x4{ 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1 };
+
+        for(size_t row = 0; row < rotationMatrix.rows; row++)
+        {
+            for(size_t col = 0; col < rotationMatrix.cols; col++)
+            {
+                transformMatrix(row, col) = rotationMatrix(row, col);
+            }
+        }
+
+        transformMatrix(0, 3) = translationVector[0];
+        transformMatrix(1, 3) = translationVector[1];
+        transformMatrix(2, 3) = translationVector[2];
+
+        return transformMatrix;
+    }
+
+    Zivid::Matrix4x4 estimateArUcoMarkerPose(const Zivid::PointCloud &pointCloud,
+                                             const std::vector<cv::Point2f> &markerCorners)
+    {
+        // Extracting 2D corners and estimateing 2D center
+        const auto center2D = estimate2DMarkerCenter(markerCorners);
+
+        // Estimating 3D corners and center from 2D data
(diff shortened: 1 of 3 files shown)


source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+} // namespace
+
+int main()
+{
+    try
+    {
+        Zivid::Application zivid;
+
+        const auto arucoMarkerFile = std::string(ZIVID_SAMPLE_DATA_DIR) + "/ArucoMarkerInCameraOrigin.zdf";
```
eskaur, line 257: Where do I get this file?

eskaur: Got it personally for testing purposes. Remember to update the official sample data zip before or right after merge.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+        std::cout << "Detecting ArUco Marker" << std::endl;
+        cv::aruco::detectMarkers(grayImage, markerDictionary, markerCorners, markerIds, detectorParameters);
+
+		if(markerIds.empty())
+        {
+            std::cout << "No ArUco markers detected";
+            exit(EXIT_SUCCESS);
+        }
+
+		std::cout << "Displaying detected ArUco marker" << std::endl;
```
eskaur, line 282: Some of your indentations are tabs. This is why it displays weird here. You can fix this by replacing tabs with four spaces.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+        detectorParameters->cornerRefinementMethod = cv::aruco::CORNER_REFINE_SUBPIX;
+
+        std::cout << "Detecting ArUco Marker" << std::endl;
+        cv::aruco::detectMarkers(grayImage, markerDictionary, markerCorners, markerIds, detectorParameters);
+
+		if(markerIds.empty())
+        {
+            std::cout << "No ArUco markers detected";
+            exit(EXIT_SUCCESS);
+        }
```
eskaur, line 280: I think you should also error out if you detect more than one marker, since then it's ambiguous what to do next. I see you currently just use the marker at index 0, but that's kind of arbitrary.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+    {
+        cv::namedWindow(bgrName, cv::WINDOW_AUTOSIZE);
+        cv::imshow(bgrName, bgr);
+        cv::waitKey(0);
+    }
+
+} // namespace
+
+int main()
+{
```
eskaur, line 252: Super nice and readable main function. Love it!

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+        cv::Ptr<cv::aruco::DetectorParameters> detectorParameters = cv::aruco::DetectorParameters::create();
+        detectorParameters->cornerRefinementMethod = cv::aruco::CORNER_REFINE_SUBPIX;
+
+        std::cout << "Detecting ArUco Marker" << std::endl;
+        cv::aruco::detectMarkers(grayImage, markerDictionary, markerCorners, markerIds, detectorParameters);
+
+		if(markerIds.empty())
+        {
+            std::cout << "No ArUco markers detected";
+            exit(EXIT_SUCCESS);
```
eskaur, line 279: Do you want to call this a "success"? Not sure about your conventions in the samples, but think about it at least.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+    Line2d fitLine(const cv::Point2d &point1, const cv::Point2d &point2)
+    {
+        // Fitting a line y=a*x + b to 2 points
+        const auto slope{ (point2.y - point1.y) / (point2.x - point1.x) };
+        const auto intercept{ point1.y - slope * point1.x };
+        return { slope, intercept };
+    };
+
+    cv::Point2f estimate2DMarkerCenter(const int &markerId,
+                                                     std::vector<cv::Point2f> &markerCorners)
```
eskaur, line 41: Make this one "const".

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+        transformMatrix(0, 3) = translationVector[0];
+        transformMatrix(1, 3) = translationVector[1];
+        transformMatrix(2, 3) = translationVector[2];
+
+        return transformMatrix;
+    }
+
+    Zivid::Matrix4x4 estimateArUcoMarkerPose(const Zivid::PointCloud &pointCloud,
+                                                  const int &markerId,
+                                                  std::vector<cv::Point2f> &markerCorners)
```
eskaur, line 147: Make this one "const".

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        transformMatrix(0, 3) = translationVector[0];
+        transformMatrix(1, 3) = translationVector[1];
+        transformMatrix(2, 3) = translationVector[2];
+
+        return transformMatrix;
+    }
+
+    Zivid::Matrix4x4 estimateArUcoMarkerPose(const Zivid::PointCloud &pointCloud,
```
eskaur, line 145: This seems like an algorithm that could easily be adapted to "estimateCheckarboardPose" instead. Is there a reason why you went for the Aruco?

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+                                                  const int &markerId,
+                                                  std::vector<cv::Point2f> &markerCorners)
+    {
+        // Extracting 2D corners and estimateing 2D center
+
```
eskaur, line 150: Unnecessary empty line here.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+        const auto center2D = estimate2DMarkerCenter(markerId, markerCorners);
+
+        // Estimating 3D corners and center from 2D data
+        const auto corners3D = estimate3DMarkerPoints(pointCloud, markerCorners);
+        const auto center3D = estimate3DMarkerPoints(pointCloud, std::vector<cv::Point2f>(1, center2D));
```
eskaur, line 155: You can write this nicer as:
```
const auto center3D = estimate3DMarkerPoints(pointCloud, { center2D });
```

eskaur: Maybe even save the one element instead of the vector:
```
const auto center3D = estimate3DMarkerPoints(pointCloud, { center2D })[0];
```
since your left hand name is singular. Then you can remove the "[0]" from the origin calculation line.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+        // Extracting origin and calcualting normal vectors for x-, y- and z-axis
```
eskaur, line 157: Typo: "calcualting"

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+            // Linear interpolation in y direction
+            // f(x,y) = (y2 - y)/(y2 - y1) * f(x,y1) + (y - y1)/(y2 - y1) * f(x,y2)
+            markerPoints3D.emplace_back(cv::Point3f{ (yRoundedUp - y) / (yRoundedUp - yRoundedDown) * fxy1.x
+                                                         + (y - yRoundedDown) / (yRoundedUp - yRoundedDown) * fxy2.x,
+                                                     (yRoundedUp - y) / (yRoundedUp - yRoundedDown) * fxy1.y
+                                                         + (y - yRoundedDown) / (yRoundedUp - yRoundedDown) * fxy2.y,
+                                                     (yRoundedUp - y) / (yRoundedUp - yRoundedDown) * fxy1.z
+                                                         + (y - yRoundedDown) / (yRoundedUp - yRoundedDown) * fxy2.z });
+        }
```
eskaur, line 121: Don't have the time to review this math. Hopefully it's from a legit source and it's tested :)

eskaur (commented):
Looks very good. I also tested actually running it, and checked the output point cloud.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: done

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: I agree with moving this check after displaying the marker. That is done.

What if a customer has a scene with two markers?
In this state of the code sample, it will work (on a random marker, yes).
If I implement that only one marker in the scene is necessary for the sample to work, then a customer needs to cover the other markers or move them out of the scene, which I don't like.

What do you think?

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: I need your input here. I don't know what is the best way to break from main here.
The first and the most basic option I tried was `return {}` but then I also thought `EXIT_SUCCESS` and `EXIT_FAILURE` might be better options. If there is no marker in the scene the code sample exits as it should, it is not unexpected behavior, which is why I thought `EXIT_SUCCESS` was a better option. Please correct me if I am wrong.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: done

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: done

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: We went for ArUco because people often use ArUco anyway. The markers can be made small and added to a scene permanently for use cases where the accuracy of detection is not too important (e.g. 3D ROI cropping, etc.)\

I also think the checkerboard pose estimation is useful and we might make a similar sample later for the checkerboard, or even better we might get it from the SDK :)

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: done

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: I think the second option (removing [0] from the origin line) would not work because I use the same function for the vector of points (4 x 2) in the above line.

Going for the first option.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: done

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: Correct. The sample is tested from a couple of different poses and it makes sense every time.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: Leaving this one unresolved as a reminder.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
eskaur: You have good points. You can leave it as is, and allow multiple markers.

Have you checked that "cv::aruco::drawDetectedMarkers" doesn't crash or something if there are no markers?

SatjaSivcev: Yes, it just returns empty as it should.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
eskaur: I agree that you should do either `return EXIT_SUCCESS` or `return EXIT_FAILURE`. (no need to use `exit()`)

Which one of the two is up to you I think. Is "no marker" a failure? Maybe, maybe not. I'm fine with either. But please change to use return-statement instead of exit().

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: I haven't done that. Good point.

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
SatjaSivcev: Talked IRL, change to return EXIT_SUCCESS

source/Applications/Advanced/TransformPointCloudViaArucoMarker/TransformPointCloudViaArucoMarker.cpp
eskaur: Tested without marker and it works fine. 
One thing. Add "endl":
```
std::cout << "No ArUco markers detected" << std::endl;
```

eskaur: Also you pushed the commented-away display line :)

SatjaSivcev: fixed 👍

eskaur (approved):

PR #169, opened 2021-09-27
Split OpenCV ArUco dependency

SatjaSivcev:


continuous-integration/linux/build.sh
```diff
@@ -8,12 +8,12 @@ BUILD_ROOT_DIR="$ROOT_DIR/build/ci"
...
 if [[ "$VERSION_ID" == "16.04" ]]; then
-    OS_SPECIFIC_OPTIONS="-DUSE_PCL=OFF -DUSE_EIGEN3=OFF -DUSE_OPENCV=OFF"
+    OS_SPECIFIC_OPTIONS="-DUSE_PCL=OFF -DUSE_EIGEN3=OFF -DUSE_OPENCV=OFF -DUSE_ARUCO=OFF"
 elif [[ "$VERSION_ID" == "18.04" ]]; then
-    OS_SPECIFIC_OPTIONS="-DUSE_EIGEN3=OFF -DUSE_OPENCV=OFF"
+    OS_SPECIFIC_OPTIONS="-DUSE_EIGEN3=OFF -DUSE_OPENCV=OFF -DUSE_ARUCO=OFF"
 elif [[ "$VERSION_ID" == "20.04" ]]; then
     EIGEN3_INCLUDE_DIR="/usr/include/eigen3"
-    OS_SPECIFIC_OPTIONS=""
+    OS_SPECIFIC_OPTIONS="-DUSE_ARUCO=ON"
```
knatten, line 16: I guess this means the OpenCV package on Ubuntu 20.04 comes with the Aruco package?

Why do you have to explicitly set `USE_ARUCO=ON` here, when you don't have to set e.g. `USE_PCL=ON` and `USE_OPENCV=ON` to use those?

knatten: Oh wait, I saw the cmake file now. This one defaults to `OFF`, but the other default to `ON`. Then you don't need to set them to `OFF` for 16 and 18. But maybe this whole code would be clearer if you explicitly set all of them for all OSes, what do you think?

knatten:
Do you know why it says "ArUco samples have been disabled" in CI for 20.04? Shouldn't they be enabled there? https://dev.azure.com/zivid-devops/zivid-cpp-samples/_build/results?buildId=6729&view=logs&j=c9a1f651-a109-592e-4363-9aac468edbba&t=9c3dffcf-480e-5a3f-69e4-a3d27b4217c6&l=8543

source/cmake/CompilerOptions.cmake
```diff
@@ -65,6 +66,7 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
             4242 # Narrowing conversions: Too strict and noisy for this code base
             4820 # The type and order of elements caused the compiler to add padding to the end of a struct
             4868 # compiler may not enforce left-to-right evaluation order in braced initializer list
+            4355 # used in base member initializer list
```
knatten, line 69: Can you include the full name of this warning here? I think this is about `this` being used in the base member initializer list, and currently it doesn't show.

knatten:
> This commit enables configuring samples that require only core OpenCV, while not configuring those that require OpenCV with extra samples.

I assume you mean "extra modules" here, not "extra samples"?

source/CMakeLists.txt
```diff
@@ -92,10 +94,18 @@ else()
 endif()
 
 if(USE_OPENCV)
-    find_package(OpenCV 4.1.0 COMPONENTS core highgui calib3d aruco)
-    if(NOT OpenCV_FOUND)
-        message(FATAL_ERROR "OpenCV not found. Please point OpenCV_DIR to the directory of your OpenCV installation (containing the file OpenCVConfig.cmake), or disable the OpenCV samples  with -DUSE_OPENCV=OFF.")
-    endif()
+	if(USE_ARUCO)
+		find_package(OpenCV 4.1.0 COMPONENTS core highgui calib3d aruco)
```
knatten, line 98: I suggest avoiding duplicating the opencv modules here and below. Or it will be easy for someone to update one and forget the other when we add a dependency on a new module. Something like this maybe?

```diff
diff --git a/source/CMakeLists.txt b/source/CMakeLists.txt
index 914ca18..a61efec 100644
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -94,14 +94,15 @@ else()
 endif()
 
 if(USE_OPENCV)
+    set(OPENCV_COMMON_MODULES core highgui calib3d)
        if(USE_ARUCO)
-               find_package(OpenCV 4.1.0 COMPONENTS core highgui calib3d aruco)
+               find_package(OpenCV 4.1.0 COMPONENTS ${OPENCV_COMMON_MODULES} aruco)
                if(NOT OpenCV_FOUND)
                        message(FATAL_ERROR "OpenCV not found. Please point OpenCV_DIR to the directory of your OpenCV installation (containing the file OpenCVConfig.cmake), or disable the ArUco samples  with -DUSE_ARUCO=OFF.")
                endif()
        else()
            disable_samples("ArUco")
-           find_package(OpenCV 4.1.0 COMPONENTS core highgui calib3d)
+           find_package(OpenCV 4.1.0 COMPONENTS ${OPENCV_COMMON_MODULES})
                if(NOT OpenCV_FOUND)
                        message(FATAL_ERROR "OpenCV not found. Please point OpenCV_DIR to the directory of your OpenCV installation (containing the file OpenCVConfig.cmake), or disable the OpenCV samples  with -DUSE_OPENCV=OFF.")
                endif()
```

source/CMakeLists.txt
```diff
@@ -92,10 +94,18 @@ else()
...
+		if(NOT OpenCV_FOUND)
+			message(FATAL_ERROR "OpenCV not found. Please point OpenCV_DIR to the directory of your OpenCV installation (containing the file OpenCVConfig.cmake), or disable the ArUco samples  with -DUSE_ARUCO=OFF.")
```
knatten, line 100: This might fail because OpenCV was not found at all, in which case it won't help to do `-DUSE_ARUCO=OFF`. But I can't really think of a better way to do it, so I think we can keep doing it like this. And then they'll get the other error message about `USE_OPENCV` when trying again which will fix it.

knatten:
I suggest squashing the commit called "Fix incorrect dependency for TransformPointCloudViaArucoMarker" into the one called "Split dependency for samples that require OpenCV with extra modules". The "Split dependency" commit currently introduces a dependency on the `aruco` module from OpenCV without documenting it until the next commit.

knatten:
Discussing the commit message about the disabled warnings offline. Review done.

continuous-integration/linux/build.sh
SatjaSivcev: Then the cmake file would have no influence on the build.sh file. Is that what you are suggesting?

SatjaSivcev:
> Do you know why it says "ArUco samples have been disabled" in CI for 20.04? Shouldn't they be enabled there? https://dev.azure.com/zivid-devops/zivid-cpp-samples/_build/results?buildId=6729&view=logs&j=c9a1f651-a109-592e-4363-9aac468edbba&t=9c3dffcf-480e-5a3f-69e4-a3d27b4217c6&l=8543

Becuase `USE_ARUCO` is OFF by default and then this happens:
https://github.com/zivid/zivid-cpp-samples/blob/44c55a6492bd9c1bf359b60568dc21765772822a/source/CMakeLists.txt#L63-L64

continuous-integration/linux/build.sh
SatjaSivcev: https://github.com/zivid/zivid-cpp-samples/pull/169/commits/44c55a6492bd9c1bf359b60568dc21765772822a

I can improve the commit later by combining it with the other one where I set these options.

source/cmake/CompilerOptions.cmake
SatjaSivcev: https://github.com/zivid/zivid-cpp-samples/pull/169/commits/47f2c0e7541410eabae854a9592eefb1538adfc7

SatjaSivcev:
> > This commit enables configuring samples that require only core OpenCV, while not configuring those that require OpenCV with extra samples.
> 
> I assume you mean "extra modules" here, not "extra samples"?

Fixed it locally. Will not push yet.

source/CMakeLists.txt
SatjaSivcev: How about I change `, or disable the ArUco samples  with -DUSE_ARUCO=OFF` to `.If you don't have OpenCV with extra modules (OpenCV_Contrib), disable the ArUco samples  with -DUSE_ARUCO=OFF`

continuous-integration/linux/build.sh
knatten: > Then the cmake file would have no influence on the build.sh file. Is that what you are suggesting?

I was just thinking it might be easier to read the `build.sh` file if all the `USE_` options are explicitly set for each OS. Then you aren't left wondering what the default is. I see you've done that in a fixup.

knatten:
> > Do you know why it says "ArUco samples have been disabled" in CI for 20.04? Shouldn't they be enabled there? https://dev.azure.com/zivid-devops/zivid-cpp-samples/_build/results?buildId=6729&view=logs&j=c9a1f651-a109-592e-4363-9aac468edbba&t=9c3dffcf-480e-5a3f-69e4-a3d27b4217c6&l=8543
> 
> Becuase `USE_ARUCO` is OFF by default and then this happens:
> 
> https://github.com/zivid/zivid-cpp-samples/blob/44c55a6492bd9c1bf359b60568dc21765772822a/source/CMakeLists.txt#L63-L64

Yes, but `build.sh` sets aruco to on for Ubuntu 20.04, right? So they shouldn't get disabled on that OS in CI?

source/CMakeLists.txt
knatten: The user could still get that error message in the case where they don't have OpenCV at all. And then turning off aruco won't help. My comment was just to acknowledge that I saw a potential small issue with this, but that I don't think we need to care about.

That being said, I think your suggested update to the error message is great!

SatjaSivcev:
> > > Do you know why it says "ArUco samples have been disabled" in CI for 20.04? Shouldn't they be enabled there? https://dev.azure.com/zivid-devops/zivid-cpp-samples/_build/results?buildId=6729&view=logs&j=c9a1f651-a109-592e-4363-9aac468edbba&t=9c3dffcf-480e-5a3f-69e4-a3d27b4217c6&l=8543
> > 
> > 
> > Becuase `USE_ARUCO` is OFF by default and then this happens:
> > https://github.com/zivid/zivid-cpp-samples/blob/44c55a6492bd9c1bf359b60568dc21765772822a/source/CMakeLists.txt#L63-L64
> 
> Yes, but `build.sh` sets aruco to on for Ubuntu 20.04, right? So they shouldn't get disabled on that OS in CI?

This (how CI works) is beyond my knowledge, unfortunatelly.

source/CMakeLists.txt
SatjaSivcev: done

source/CMakeLists.txt
SatjaSivcev: done

SatjaSivcev:
Ready for the next round @knatten

knatten:
> This (how CI works) is beyond my knowledge, unfortunatelly.

I'll investigate.

knatten:
> > This (how CI works) is beyond my knowledge, unfortunatelly.
> 
> I'll investigate.

It's from linting. Aruco samples get disabled by default, and linting doesn't explicitly enable or disable and samples. So currently we aren't linting Aruco samples. I'll push a commit that fixes it.

knatten:
> > > This (how CI works) is beyond my knowledge, unfortunatelly.
> > 
> > 
> > I'll investigate.
> 
> It's from linting. Aruco samples get disabled by default, and linting doesn't explicitly enable or disable and samples. So currently we aren't linting Aruco samples. I'll push a commit that fixes it.

I pushed a fixup for your initial commit on this branch now. Also took the liberty to sneak in another tiny commit which slightly improves the linter output, I hope you agree with it?

continuous-integration/linux/lint.sh
```diff
@@ -41,6 +41,7 @@ cmake -GNinja \
     -DCMAKE_CXX_CLANG_TIDY="/usr/bin/clang-tidy-10" \
     -DWARNINGS=ON \
     -DWARNINGS_AS_ERRORS=ON \
+    -DUSE_ARUCO=ON \
```
SatjaSivcev, line 44: Should we set all of them on for linting?

SatjaSivcev: I can do it. What do you think? I think it makes sense. Then, in future, if we disable some by default, the liner will do its job anyway for all the files.

knatten: Yeah, good point! I should have thought of that since I was the one suggesting it for `build.sh` :grinning:

SatjaSivcev: done

SatjaSivcev: https://github.com/zivid/zivid-cpp-samples/pull/169/commits/61123d530ec7c7034e74b4893cbe5f2178dd22cf

I added "lint" to the commit message.

SatjaSivcev:
> > > > This (how CI works) is beyond my knowledge, unfortunatelly.
> > > 
> > > 
> > > I'll investigate.
> > 
> > 
> > It's from linting. Aruco samples get disabled by default, and linting doesn't explicitly enable or disable and samples. So currently we aren't linting Aruco samples. I'll push a commit that fixes it.
> 
> I pushed a fixup for your initial commit on this branch now. Also took the liberty to sneak in another tiny commit which slightly improves the linter output, I hope you agree with it?

squashed

SatjaSivcev:
> Discussing the commit message about the disabled warnings offline. Review done.

I found out what commits introduced the warnings and I added that to the commit message that ignores them.

knatten (approved):
Good stuff!

PR #187, opened 2021-11-19
Update CMakeLists

SatjaSivcev:
This commit does blah blah.


source/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
-cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.3 FATAL_ERROR)
```
SatjaSivcev, line 1: This is not ok, it should stay 3.5

PR #195, opened 2022-01-25
Various sample updates

SatjaSivcev:
New samples
- Adds TransformPointCloudFromMillimetersToMeters sample.
  This sample transforms   a point cloud from millimeters
  to meters using transform API.

Modifications
- Adds save to CaptureUndistort sample
- Adds export to ply to Capture sample
- Adds save and load settings to CaptureHDRCompleteSettings sample
- Adds downsample as a new point cloud instance in addition to the
  in-place in downsampling sample
- Updates UtilizeEyeInHand sample and renames it to UtilizeHandEye.
  This sample now works with a new dataset for eye-in-hand
  (located in SampleData.zip), consisting of:
  - RobotTransform
  - EyeInHandTransform
  - ZividGemEyeInHand
  - ZividGemEyeToHandTransformed
  and the following dataset for eye-to-hand, consisting of:
  - ZividGemEyeToHand
  - ZividGemEyeToHandTransformed
  The sample now has the following options:
  - eye-in-hand or eye-to-hand
  - transform picking point or entire point cloud
  - renames all the samples to UtilizeHandEyeCalibration

General changes
- Improves C++ code in general
- Improves descriptions and printouts


.gitattributes
```diff
@@ -1 +1,2 @@
+* text=auto
```
chrisasc, line 1: do you know what this do?

chrisasc (approved):
lgtm, I only have a small questions out of curiosity.

.gitattributes
SatjaSivcev: From the docs for the text setting:

`This attribute enables and controls end-of-line normalization. When a text file is normalized, its line endings are converted to LF in the repository.`

And auto means that 

`the path is marked for automatic end-of-line conversion. If Git decides that the content is text, its line endings are converted to LF on checkin.`

So whenever you commit a text file, the line endings are converted to LF no matter if you're on Windows or Linux.

